PR #17, opened 2026-08-17
B1 spike: does a BAML client fit ARGUS's LLM transport?

abhisheksharma001:
## What & why

The BAML adoption tracker gates everything on **B1**:

> Does BAML's client work against the Supabase edge function shape, or does the edge function need a compatibility route first?

**It works as-is. No change to `index.ts` is needed.** That unblocks the rest of the tracker.

This PR adds the spike that establishes it — `scripts/baml_spike/`, nine probes against a local replica of the transport. No API keys, no Supabase project, no network:

```bash
pip install baml-py==0.225.0
python scripts/baml_spike/run_spike.py --verbose
```

`fake_proxy.py` reproduces `supabase/functions/llm-proxy/index.ts` path by path, including the parts that differ from OpenAI: the bare `{"error": "..."}` failure envelope, the 200-calls/day 429, and the fact that the function never inspects the request path. Full writeup in [`FINDINGS.md`](https://github.com/VaradDurge/ARGUS/blob/claude/baml-adoption-tracker-rpgadi/scripts/baml_spike/FINDINGS.md).

| # | Question | Verdict |
|---|---|---|
| P1 | Can a BAML client reach the edge function at all? | ✅ PASS |
| P2 | Does BAML's auth header match what `index.ts` expects? | ✅ PASS |
| P3 | Do `max_tokens` / `temperature` / `response_format` survive to the wire? | ⚠️ CAVEAT |
| P4 | What happens to the bare `{"error": "..."}` envelope? | ⚠️ CAVEAT |
| P5 | Can a refreshed Supabase JWT be injected per call? | ✅ PASS |
| P6 | Is BAML's parser better than `_extract_json_object`? | ⚠️ CAVEAT |
| P7 | Can the per-call timeouts be preserved? | ⚠️ CAVEAT |
| P8 | Does a `retry_policy` amplify load on the hot path? | ⚠️ CAVEAT |
| P9 | Can one BAML function cover the BYOK provider matrix? | ⚠️ CAVEAT |

### The three findings that change scoping

**Per-call timeouts have no declarative support (P7).** There is no `timeout` in `BamlCallOptions`. `timeout_ms` and `request_timeout_ms` *are* accepted as client options, which makes them look like the answer — but BAML forwards them to the provider as request body fields rather than enforcing them client-side. Against a server hanging 3s with `timeout_ms=500` set, the call blocked the full 3s. What works is `AbortController` + a caller-side timer, which fired at exactly 1.00s. So the tracker's 5.0s / 8.0s / 15.0s / 30.0s are preservable, but as a timer+abort wrapper at every call site rather than a BAML policy.

**A `retry_policy` turns one logical call into three requests (P8).** BAML treats a 429 as transient. The proxy's 429 is a hard daily quota (`DAILY_LIMIT = 200`), so retrying spends quota that is already gone and triples the latency of a call that was always going to fail. Worst exactly on the hot path the tracker proposes piloting. Retries are one of the three things listed as BAML's value proposition; on the proxy path the default behaviour is a liability.

**The proxy is path 2 of 2 (P9).** `create_chat_completion` now tries BYOK *first* across `openai | anthropic | google` and only falls back to the proxy — for any user who has run `argus key set`, the proxy is never taken. `providers.py` does real per-provider translation that `openai-generic` wouldn't reproduce (`_split_system()` into Anthropic's `system` and Gemini's `systemInstruction`, `contents`/`parts`, `maxOutputTokens`, `responseMimeType`, three response shapes normalized by `_envelope()`). The unit of migration is not "six call sites" but six sites × two transport paths × three providers.

### Packaging numbers

Eight `abi3` wheels (`cp38`), so one wheel per platform covers the whole 3.9–3.12 matrix with no per-version builds. ~21 MB each, ~54 MB installed. **No sdist** — any platform outside that list has no install path at all, which narrows support from "anywhere Python runs" (three pure-Python deps today, stdlib `urllib` for every LLM call) to that table. `baml-cli generate` runs in 0.08s and ships with `baml-py`, so the CI codegen step is cheap — not the blocker.

## Type of change

- [x] Other: spike / investigation — no production code touched

## Checklist

- [x] My change targets the **open-source core** only — `cloud/` and `supabase/` are untouched. `fake_proxy.py` *replicates* the edge function contract for testing but modifies nothing under `supabase/`.
- [x] `pytest` passes locally — 574 passed, 3 skipped, 2 xfailed.
- [x] `ruff check .` passes.
- [ ] Docs updated — n/a, no behavior or public API change.

## Notes for reviewers

**Scope.** Spike only. Adds no dependency to `pyproject.toml`, imports nothing from `argus`, is imported by nothing in `argus`, and ships in neither distribution (`scripts` is already in the sdist exclude list; the wheel only packages `src/argus`). Deleting the directory is a clean no-op. `baml_client/` is generated and gitignored — `run_spike.py` regenerates it, which keeps CI's `ruff check .` off 1300 lines of generated code and doubles as evidence for the codegen question.

**One live bug found, deliberately not fixed here.** P6 feeds a judge response of `{"reason": "ok"}` — no verdict at all — through the current parser. `bool(parsed.get("pass", True))` returns a **passing** `SemanticCheckResult` with `evaluated=True`, so a malformed response is indistinguishable from a genuine pass. Every other parse failure in that function routes to `_skip_result(...)` with `evaluated=False`, correctly read as "no opinion"; this path skips that. Needs no dependency and is worth fixing whether or not BAML is adopted — kept out of this PR so a spike doesn't quietly change judge verdict behavior. Happy to send it as its own PR.

**Suggested input to B3.** The tracker picks `semantic_checker.py` as pilot on blast-radius and hot-path grounds, both still valid. But P6 shows it is the one site whose parsing BAML barely improves — it already hand-rolls fence-stripping and truncation repair — and P7/P8 show the hot path is where BAML's timeout and retry gaps hurt most. A pilot there measures BAML at its least favourable. `heuristic_disambiguator.py` (the tracker's own alternative) is on bare `json.loads`, batches one call per node, and has an 8.0s timeout. Trades a better measurement for slightly more blast radius. Your call — flagging, not deciding.

**Why a replica instead of live Supabase.** B1 is a question about wire contracts, and a replica pins those exactly while staying reproducible on any branch without a login or quota. Two things it can't answer, both out of scope: whether real Supabase edge routing rejects the `/chat/completions` subpath before the function runs (it routes on function name, so it shouldn't — worth one live call to confirm before the pilot), and how real models behave under BAML's prompt-embedded schema versus OpenAI JSON mode, which is a prompt-quality question for the pilot rather than a transport one.

---


abhisheksharma001:

VaradDurge (dismissed):
Looks good, nice work. Fake proxy matches index.ts faithfully — checked the auth parsing, error envelope shape, and path-ignoring behavior against the real function.

P7 timeout finding is the sharpest one here, glad you dug into that instead of trusting the option name at face value. P8 retry + hard quota interaction is a good catch too.

Two things:

1. Please send the parsed.get("pass", True) bug fix as its own PR like you offered — that's worth fixing regardless of where BAML lands.

2. Minor nit: fake_proxy doesn't replicate the defaults index.ts applies before forwarding (model ?? "gpt-4o", temperature ?? 0.3, max_tokens ?? 2000). Doesn't affect any probe conclusions since you're measuring what BAML sends not what OpenAI receives, but a one-line comment noting the simplification would help future readers.

Agree on reconsidering the pilot target for B3 — your reasoning on heuristic_disambiguator makes sense.

abhisheksharma001:
Thanks for the review. All three addressed, but please read the third one before acting on the B3 agreement.

**1. The `parsed.get("pass", True)` fix — sent as #19.**

It went in on its own branch off master, no BAML dependency. The scope turned out wider than this spike documented, so flagging that rather than letting it land quietly: tracing the consumer chain showed the failure has teeth beyond a mislabelled skip. `session._apply_judge_verdict` acts on `passed=True` + `evaluated=True` + `confidence >= 0.7` by flipping a heuristic failure to `pass` **and** writing an `llm_full_override` into the feedback store that feeds the auto-approve threshold. A response like `{"reason": "looks fine", "confidence": 0.95}` — no verdict at all — clears both gates. Run against master's module and the fixed one:

```
BEFORE  passed=True  confidence=0.95  evaluated=True   -> heuristic 'fail' becomes: 'pass'
AFTER   passed=True  confidence=0.0   evaluated=False  -> heuristic 'fail' becomes: 'fail'
```

The `{"reason": "ok"}` case I documented here is the benign end — no `confidence`, so it misses the 0.7 gate. Two further routes reach the same fabrication, and #19 covers all three; the sharpest is that `bool("false")` is `True`, so a judge explicitly failing a node under JSON mode was recorded as passing it.

**2. The `fake_proxy` nit — fixed in fd743b5.** Note in the module docstring plus an inline comment at the forward point. I framed it as deliberate rather than a shortcut, since P3 actually depends on it: the probe measures which fields BAML puts on the wire, so a field BAML omits has to stay omitted in the replica to be observable. Filling in `model ?? "gpt-4o"` / `max_tokens ?? 2000` / `temperature ?? 0.3` would mask the thing P3 is looking for.

**3. The B3 recommendation you agreed with is wrong, and I'm retracting it — see #20.**

That PR went up a couple of minutes before your review, so the timing is unlucky. `heuristic_disambiguator.py` is unreachable. Nothing in the package imports it, from either entry point — `argus` or `argus.cli.main`. A pilot there would measure nothing, and the reasoning you agreed with (bare `json.loads`, 8.0s timeout, batches one call per node) describes a code path that never executes.

It is worse than one bad pick. Checking the whole inventory rather than the six names in §2 turned up seven LLM call sites, three unreachable:

```
argus.heuristic_disambiguator      1x      UNREACHABLE  shared proxy    yes
argus.llm_correlator               1x      UNREACHABLE  shared proxy    yes
argus.llm_investigator             3x      reachable    shared proxy    yes
argus.loop_analyzer                1x      UNREACHABLE  shared proxy    MISSING
argus.semantic_checker             1x      reachable    shared proxy    yes
argus.signature_generalizer        1x      reachable    direct OpenAI   yes
argus.source_locator               1x      reachable    shared proxy    yes
```

`loop_analyzer.py` is a seventh site §2 never listed — same proxy call, same hand-parse, docstring opens "Mandatory, always-on analysis for looped nodes", never called. And half of §4's timeout table belongs to dead modules: the 8.0s and 15.0s rows are the two unreachable ones, which shrinks the P7 abort-wrapper problem you singled out to a single live override rather than a per-site matrix.

§5's only two candidates were `heuristic_disambiguator` and `semantic_checker`, so B3's menu is down to one item unless the ranking reopens. #20 reopens it over the four live sites and lays out the tradeoff — `semantic_checker` for a low-risk integration proof, `llm_investigator` if the pilot is meant to answer whether BAML earns its place. I did not pick between them; they answer different questions and the tracker hasn't said which one it's asking.

---

PR #18, opened 2026-08-17
B2: route signature generalization through the shared LLM path

abhisheksharma001:
## What & why

Answers **B2** from the BAML adoption tracker:

> Does `signature_generalizer.py` move onto the proxy?

**Yes — the chat call does, and it should regardless of BAML.** `_llm_generalize` was the last LLM call site in the package still holding its own transport:

```python
from argus.embedding_store import _get_client
client = _get_client()
...
resp = client.chat.completions.create(model="gpt-4o-mini", ...)
```

`_get_client()` resolves `resolve_openai_key()` only. So for anyone on Anthropic or Google, this call could never succeed — it raised, got swallowed by `except Exception: return None`, and every signature quietly fell back to the heuristic generalizer. No warning, no `error` field, nothing in the run record. The failure was indistinguishable from "the LLM produced a bad regex."

This moves it to `llm_proxy.create_chat_completion`, the same entry point `semantic_checker`, `heuristic_disambiguator`, `llm_correlator`, `llm_investigator`, `loop_analyzer`, and `source_locator` already use. That resolves BYOK across all three providers first and falls back to the hosted proxy, so generalization now works for:

- Gemini and Claude users (previously: heuristic only)
- logged-in users with no key at all (previously: heuristic only)
- OpenAI users — unchanged, `resolve_openai_key()` is literally `resolve_key("openai")`, and `resolve_model()` passes OpenAI model names straight through

`providers.resolve_model()` maps the `gpt-4o-mini` tier hint to `claude-3-5-haiku-latest` / `gemini-2.5-flash` for the other two.

### What's preserved

The fail-open contract, unchanged: every failure path still returns `None` and the caller still falls back to `_heuristic_generalize`. The returned regex is still validated twice — it must compile, and it must match all of its own evidence strings, or it's discarded as too narrow.

### Three behavior changes worth reviewing

**Errors are now checked explicitly.** `create_chat_completion` reports failure in-band as `{"error": "..."}` rather than raising, so relying on the `except Exception` clause would have silently accepted an error dict as a response. There's now an explicit `if "error" in result` check, plus a guard for a missing/empty `choices` list.

**The call is bounded at 15s.** It previously passed no timeout and inherited the OpenAI SDK's 600s default — inside `session.py`'s loop over `suggested_signatures`, once per signature at session end. 15s is not one of the tracker's §4 hot-path budgets (this isn't per-node); it's sized for a ≤300-token completion with headroom while keeping N signatures from stalling a run finalize.

**An empty completion is now rejected.** This is a real bug in the current code, on the line this PR rewrites. `""` passes *both* existing validations — `re.compile("")` succeeds, and `re.compile("").search(anything)` matches — so an empty model response produced a stored signature whose pattern matched every output of every node. The old `.strip()` on `resp.choices[0].message.content` raised on `None` and so was safe, but an empty *string* was not. Now guarded.

### What deliberately does not move

`cluster_with_existing()` stays on `embedding_store`. It needs **embeddings**, and there is no embeddings transport to move it to — `llm_proxy` exposes chat completions only, and neither the Anthropic nor the Google adapter in `providers.py` implements an embeddings endpoint. Candidate clustering therefore still requires an OpenAI key specifically. That's a separate transport question (a proxy `embeddings` route, or per-provider embedding adapters) and is not in B2's scope — flagging it so it doesn't get lost.

## Type of change

- [x] Bug fix — non-OpenAI users silently got no LLM generalization; empty completions produced match-everything patterns
- [x] Refactor / internal — last call site onto the shared transport

## Checklist

- [x] My change targets the **open-source core** only — I have **not** modified `cloud/` or `supabase/`.
- [x] Tests added/updated and `pytest` passes locally — 585 passed, 3 skipped, 2 xfailed (11 new).
- [x] `ruff check .` passes.
- [ ] Docs updated — n/a, no public API or documented behavior change.

## Notes for reviewers

**Test coverage.** The 11 new tests are in a `LLM transport tests` section and mock `argus.llm_proxy.create_chat_completion` / `is_available`, matching the idiom in `test_heuristic_disambiguator.py`. They cover the success path, the `{"error": ...}` path, an exception raise, `is_available() == False`, code-fence stripping, a too-narrow regex, the empty-response guard, a missing `choices` list, the model/timeout/`max_tokens` arguments, and one end-to-end assertion that the LLM regex beats the heuristic. One of them (`test_llm_generalize_never_touches_embedding_store_client`) monkeypatches `embedding_store._get_client` to raise, so it fails loudly if this call site ever regains its own client.

**The pre-existing tests are unchanged.** All of them patch `_llm_generalize` wholesale, so they never exercised the transport — which is how this could sit broken for non-OpenAI users without a red test.

**No `response_format`.** The prompt asks for a bare regex string, not JSON, so nothing is imposed. Whether a Gemini or Claude model reliably returns a bare regex rather than prose is a prompt-quality question this PR doesn't settle — but the failure mode is benign: prose either fails `re.compile` or fails the evidence check, and the caller lands on the heuristic, which is exactly what those users get today.

**Relationship to #17.** Independent — that PR only adds `scripts/baml_spike/`, this one only touches `src/argus/`. Both branch from `master` and can merge in either order. Nothing here depends on BAML being adopted; B2 was worth doing on its own terms and this is the answer either way.

---


VaradDurge (approved):
Solid fix. Checked the diff against the old _get_client() path — the silent failure for non-OpenAI users was real and this is the right fix.

The empty-pattern guard is a good catch, re.compile("") matching everything is nasty. And the explicit error/choices checks are the right thing to do since create_chat_completion reports errors in-band.

Tests are thorough, especially the monkeypatch on _get_client that'll blow up if someone accidentally reintroduces the direct client.

One question for my own understanding: for the prompt asking for a bare regex (no response_format), have you seen any cases where Gemini or Claude return prose instead of a pattern? The fail-open path handles it fine either way, just curious if you tested it live against non-OpenAI models or only through the mocks.

LGTM, ship it.

abhisheksharma001:
Straight answer: **only through mocks.** I don't have Gemini or Anthropic keys in this environment, so this has never run against a live non-OpenAI model. Everything below comes from probing the validation path, not from observed traffic — treat it as "what the code does with prose," not "what those models actually return."

What I could check is which prose shapes survive the compile + evidence checks. Six shapes through the real `_llm_generalize`:

```
a real regex (control)             ACCEPTED
prose prefix + regex               REJECTED -> heuristic
pure explanation                   REJECTED -> heuristic
prose w/ unbalanced paren          REJECTED -> heuristic
model refuses ("I'm sorry, ...")   REJECTED -> heuristic
model echoes the input verbatim    ACCEPTED
```

So the fail-open path holds for prose that *doesn't* match the evidence — which is nearly all of it, since explanation text has to match the evidence string to get through, and it doesn't.

**The one that survives is the echo**, and it isn't entirely benign. An echoed pattern is now compiled as a regex where it used to be a `contains_ci` literal, so metacharacters change meaning:

```
'rate limit 5.0 exceeded'  -> ACCEPTED, and also matches 'rate limit 5X0 exceeded'
'cost was $5 total'        -> REJECTED  ($ can't match mid-string, fails evidence)
'failed (timeout)'         -> REJECTED
```

`$` and unbalanced `(` get caught by the compile/evidence checks. `.` doesn't — it matches itself *plus* anything else, so the stored signature is quietly broader than the literal it came from, and gets marked `generalized=True` having generalized nothing.

To be clear about ownership: **this is pre-existing, not something this PR introduces.** Both validations are unchanged here. But this PR does increase exposure to it, and that's fair to name — it puts Gemini and Claude users on the LLM path for the first time, on the cheap tier, which is where an echo is most likely.

Cheap guard if you want it, in `_llm_generalize` next to the empty-pattern check:

```python
if regex.strip().lower() == pattern.strip().lower():
    return None  # generalized nothing; heuristic may do better
```

Happy to add it here, or file it separately so I don't reset your approval — your call. Left it out for now since you've already signed off and it's a different bug from the one B2 was about.

---

VaradDurge:
Good catch on the echo case. Yeah go ahead and add that guard here — it's one line and this PR is the one that puts more users on the LLM path, so makes sense to ship it together rather than as a separate thing. The approval can be re-done, no big deal.

Also fair point on the mocks-only testing. Not a blocker since the fail-open path handles it, but would be nice to get one live run against a non-OpenAI model before we lean on this for real. Low priority though.

PR #19, opened 2026-08-17
Require an explicit verdict from the semantic judge

abhisheksharma001:
## What & why

`check_semantic_coherence()` reported a **fabricated verdict as a real one**.

```python
passed=bool(parsed.get("pass", True)),
```

If the judge's JSON parses but carries no verdict, this returns a **passing** `SemanticCheckResult` with `evaluated=True` — indistinguishable from a genuine pass. Every other parse failure in that function routes to `_skip_result(...)` with `evaluated=False`, which callers correctly read as "no opinion". This one path skipped that.

This was flagged in #17 as found-but-not-fixed. Sending it separately: it needs no dependency and is worth having whether or not BAML is adopted.

### It has teeth

`session._apply_judge_verdict()` overrides a heuristic failure to `"pass"` when `passed` is True and `confidence >= 0.7`, and records an `llm_full_override` in the feedback store that feeds the auto-approve threshold.

A body like `{"reason": "looks fine", "confidence": 0.95}` — **no verdict at all** — met both conditions. Verified against the real `_apply_judge_verdict`, same input both sides:

```
BEFORE  passed=True  confidence=0.95  evaluated=True   -> heuristic 'fail' becomes: 'pass'
AFTER   passed=True  confidence=0.0   evaluated=False  -> heuristic 'fail' becomes: 'fail'
```

So it silently cleared a failing node *and* trained the override machinery on a ruling that was never given. The `{"reason": "ok"}` case from #17 was the benign end of the same bug — it defaults `confidence` to 0.0, misses the `>= 0.7` gate, and only mislabels coverage.

### Three routes produced a fabricated verdict

| Response | Was | Now |
|---|---|---|
| `{"reason": "ok"}` — no `pass` key | evaluated pass | skip |
| `"choices": []` — no completion returned | evaluated pass¹ | skip |
| `{"pass": "false"}` — non-boolean | evaluated **pass** | evaluated **fail** |

¹ empty `choices` was turned into a literal `"{}"`, which parses to an empty dict and takes the same `get("pass", True)` default.

The third is the sharpest. `bool()` is the wrong reader for this field — `bool("false")` and `bool("no")` are both `True`, so a judge that *explicitly failed* a node under JSON mode was recorded as passing it. `_coerce_verdict()` reads real booleans plus the string forms models emit, and returns `None` for anything else instead of guessing.

## Type of change

- [x] Bug fix

## Checklist

- [x] My change targets the **open-source core** only — I have **not** modified `cloud/` or `supabase/`.
- [x] Tests added/updated and `pytest` passes locally — 589 passed, 3 skipped, 2 xfailed (574 before, 15 new).
- [x] `ruff check .` passes.
- [ ] Docs updated — n/a. No public API change; `SemanticCheckResult` is unchanged and the `evaluated` docstring already describes this contract.

## Notes for reviewers

**Nothing becomes newly blocking.** `_skip_result` returns `passed=True`, so the three routes above go from "passing and evaluated" to "passing and not evaluated". The only status change is the one shown above — a heuristic verdict that used to get overridden now stands, which is the bug being fixed.

**One number moves.** `session.py:234` computes judge coverage as evaluated-over-judged. These responses used to count as evaluated and no longer do, so reported coverage drops slightly on runs that hit them. That is the metric being corrected, not a regression — it was counting non-verdicts as judgments.

**Scope, and where I went past what #17 documented.** The spike found route 1. Routes 2 and 3 are the same defect — fabricating a verdict the judge didn't give — reached differently, and I found them tracing route 1. Fixing route 1 alone would have left `{"pass": "false"}` inverting an explicit failure, which is worse than the bug in the ticket. Flagging the widened scope rather than burying it.

**A verdict-less response drops its disambiguation verdicts too.** Deliberate: every other skip path in this function returns `[]`, and trusting the per-signal rulings of a response too malformed to state a top-level verdict is the kind of half-trust that gets confusing later. Dropping them is also the safe direction — ambiguous signals keep their heuristic confidence, which is the pre-LLM behavior. Easy to decouple if you'd rather keep them.

**`is_failure` uses the same reader with a different default.** The prompt tells the judge "when in doubt, mark is_failure=true", so an uninterpretable value there keeps `True` rather than skipping the signal. Only the `bool("false") == True` inversion is fixed.

**Test coverage.** 15 tests: `_coerce_verdict` in isolation, the three fabrication routes end to end, string-boolean verdicts in both directions, a guard that genuine `True`/`False` rulings still evaluate, and the disambiguation cases. `ambiguous_signals` had no test coverage at all before this.

**How I verified the bug rather than just describing it.** Loaded `origin/master`'s `semantic_checker.py` alongside the fixed one and ran the same mocked response through the real `ArgusSession._apply_judge_verdict` on both — that's where the BEFORE/AFTER table above comes from.

---


abhisheksharma001:
Pushed `2315a79` after re-reviewing this branch — it resets the approval above, which is worth it because the first commit had a regression in it.

**`_coerce_verdict` skipped integer verdicts, and `bool(parsed.get("pass", True))` did not.** A judge emitting `{"pass": 0}` was an evaluated failure on master and became a skip here. This PR is about not fabricating verdicts, and it was quietly discarding a real one — in the direction that matters, since a dropped failure surfaces as a silent pass.

The table in the description above should read:

| Response | master | first commit | now |
|---|---|---|---|
| `{"pass": 0}` | evaluated **fail** | skip — verdict lost | evaluated **fail** |
| `{"pass": 1}` | evaluated pass | skip — verdict lost | evaluated pass |
| `{"pass": "0"}` | evaluated **pass** (inverted) | skip | evaluated **fail** |

So the fix restores what master got right without giving up what this PR corrected.

Only `0` and `1` qualify. The `isinstance(value, int)` guard excludes floats on purpose — `1.0 == 1` in Python, so an equality test would have let `1.0` through, and a float in a verdict field reads as a confidence that landed in the wrong key rather than a boolean the model got slightly wrong. The `isinstance(value, bool)` branch still runs first, so real booleans never reach the integer path.

Same reader still backs `is_failure`, so an explicit `0` there now clears the signal while a missing verdict keeps the prompt's documented `is_failure=true` default.

596 passed / 3 skipped / 2 xfailed (up 6), `ruff check .` clean.

---

VaradDurge (approved):
Verified — full test suite passes (627 passed, 0 failures) on this branch. The three fabrication routes are properly closed:

1. Missing `pass` key → skip (was: fabricated pass)
2. Empty `choices` → skip (was: fabricated pass via `{}` → `get("pass", True)`)
3. `"false"` string → evaluated fail (was: `bool("false") == True` → fabricated pass)

`_coerce_verdict()` is clean — handles real booleans, string forms, 0/1 integers, and rejects everything else as "no verdict". Disambiguation verdicts use the same coercion with the correct conservative default.

No existing tests broke, no API surface changed. Ship it.

PR #20, opened 2026-08-17
B3: three of the seven LLM call sites are unreachable

abhisheksharma001:
## What & why

The BAML tracker's §2 lists six modules that make an LLM call, calls the list verified, and B3 goes straight to picking one to pilot a migration on. Before answering B3 I checked the assumption underneath it: that every candidate runs.

**There are seven LLM call sites, not six, and three of them are unreachable.**

```
package modules   : 54
reachable         : 51
entry points      : argus, argus.cli.main
LLM call sites    : 7 discovered, 6 in tracker §2

module                             calls   import       transport       §2
----------------------------------------------------------------------------------
argus.heuristic_disambiguator      1x      UNREACHABLE  shared proxy    yes
argus.llm_correlator               1x      UNREACHABLE  shared proxy    yes
argus.llm_investigator             3x      reachable    shared proxy    yes
argus.loop_analyzer                1x      UNREACHABLE  shared proxy    MISSING
argus.semantic_checker             1x      reachable    shared proxy    yes
argus.signature_generalizer        1x      reachable    shared proxy    yes
argus.source_locator               1x      reachable    shared proxy    yes
```

Nothing in the package imports the three marked UNREACHABLE, from either entry point the distribution exposes — `argus` for library users, `argus.cli.main` for the console script. They cannot execute.

**This retracts a recommendation I made in the B1 spike (#17).** That spike argued for switching the pilot to `heuristic_disambiguator.py`, on the grounds that it would show BAML's parsing win more honestly than `semantic_checker.py`. It would show nothing — the module never runs. §5's two candidates were that module and `semantic_checker`, so B3's menu is down to one item unless the ranking is reopened.

### The site §2 misses

`loop_analyzer.py` (201 lines) calls the shared proxy with `response_format={"type": "json_object"}`, `json.loads`-es the reply, and defaults every field on a parse failure — the same shape as the other six. Its docstring opens *"Mandatory, always-on analysis for looped nodes."* It is never called. `RunRecord` carries a `loop_analyses` field and `storage.py` has a deserializer for it, so persisted runs read the field back; nothing writes one, so it is empty on every run.

It also carries the bug class fixed in #19, dormant: an unparseable reply becomes `data = {}` becomes a `LoopAnalysisResult` with `summary=""`, `is_stalled=False` and **`error=None`** — a clean analysis, reported as if the judge had run. Harmless while the code is dead; worth knowing before anything wires it back in.

### Two consequences for scoping

- **Neither of §4's explicit timeout overrides is on live code.** The 8.0s and 15.0s rows are `heuristic_disambiguator` and `llm_correlator`, both dead. Two live sites do pass an explicit `timeout=` — `semantic_checker` (5.0s, or 8.0s when disambiguating) and `signature_generalizer` (15.0s, which #18 added and which coincidentally matches the dead `llm_correlator` number). `llm_investigator` and `source_locator` take the 30.0s default. So the `threading.Timer` + `AbortController` wrapper the spike called its sharpest constraint has to reproduce **two** overrides plus a default, not §4's four-value matrix.
- **Every unreachable module in the package is an LLM call site.** There are exactly three, and all three are here. The dead code in ARGUS is concentrated entirely in the area the tracker is about to migrate.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / internal
- [x] Docs
- [x] Other: analysis tooling under `scripts/` — nothing under `src/` changes

## Checklist

- [x] I have signed the [CLA](../CLA.md) (the bot will prompt on first PR).
- [x] My change targets the **open-source core** only — I have **not** modified `cloud/` or `supabase/` (proprietary; maintainer-only).
- [x] Tests added/updated and `pytest` passes locally. *(585 passed, 3 skipped, 2 xfailed — unchanged from master after merging #18 in; this PR adds no runtime code. `check_inventory.py` is analysis tooling in the same category as `run_spike.py`, which CI does not run either.)*
- [x] `ruff check .` passes.
- [x] Docs updated if behavior or the public API changed. *(No behavior change. `INVENTORY.md` records the run.)*

## Notes for reviewers

**How the seventh site surfaced.** `check_inventory.py` discovers call sites by scanning every module for `create_chat_completion(...)` and `...chat.completions.create(...)`, then diffs the result against §2 — rather than reading the list off the tracker and checking those six. `loop_analyzer` appeared in that diff. The check also catches the reverse case (a listed module with no LLM call).

**Why grep alone wasn't enough.** Two things flip the answer if you skip them, and I hit both while writing this:

- *Function-local imports.* ARGUS uses them throughout to keep optional paths cheap. A graph ignoring them reports most of the package as dead.
- *Ancestor packages.* Importing `argus.utils.serializer` executes `argus/utils/__init__.py`. My first version omitted this and reported `argus.utils` as unreachable, which is wrong. Caught by cross-checking every verdict against grep.

Every claim in `INVENTORY.md` has a file:line behind it — `--verbose` prints the shortest import path to each site and each call line.

**Transport is detected, not hardcoded** — read from imports rather than a fixed list. #18 has since merged, and the checker picked the change up on its own: `signature_generalizer` now reports `shared proxy` where it reported `direct OpenAI` when this PR was opened, with no edit to the tool. The prose around it did *not* self-correct, and `4d1443c` fixes the four statements #18 invalidated (the pasted run, the §4 table row, "the only live site that passes a `timeout=`", and "one explicit override"). That gap is worth naming on its own: the measurement tracked master and the conclusions drawn from it did not.

**Three things I deliberately did not do:**

1. **Delete the dead modules.** 567 lines plus a test file, and they are a legal (if undocumented and unexported) import for a library consumer. Maintainer's call. `loop_analyzer` is the interesting one — its storage and model plumbing already exist, and "mandatory, always-on" reads like intent that was never finished, so wiring it back in is as plausible as deleting it.
2. **Fix the dormant bug in `loop_analyzer`.** Fixing code that cannot run is churn; the decision above comes first.
3. **Pick the pilot.** `INVENTORY.md` ranks the four live sites and lays out the tradeoff — `semantic_checker` for a low-risk integration proof, `llm_investigator` if the pilot is meant to answer whether BAML earns its place. Those answer different questions and the tracker has not said which one it is asking. Flagging, not deciding.

**Blast radius:** none at runtime. Two new files under `scripts/baml_spike/`; no imports, no dependencies beyond the standard library.

**Relationship to the rest of the series.** #17 (B1) and #18 (B2) have both merged; master is merged into this branch. `INVENTORY.md` links to `FINDINGS.md` and restates the claims it borrows, so it still reads standalone.

**Why `tests/AUDIT_REPORT.md` didn't catch this:** `heuristic_disambiguator` has 7 passing tests and is recorded at 93% coverage. The tests exercise the module directly, so they stay green while nothing else calls it. Coverage measures whether a line ran under test, not whether production reaches it.


VaradDurge (approved):
CI green across all three Python versions (3.9, 3.11, 3.12). No runtime code touched — changes are entirely under scripts/baml_spike/.

Good find on the three unreachable LLM call sites and the missing loop_analyzer from the tracker. The check_inventory.py approach (discover-then-diff rather than check-a-list) is the right one — it caught what grep alone would miss (function-local imports, ancestor packages).

INVENTORY.md is thorough and the scoping consequences for B3 are clearly laid out. Ship it.

PR #29, opened 2026-08-19
feat: Phase 1 — Time-Travel State Patching (steps 1.1–1.4)

abhisheksharma001:
## What this adds

`ReplayEngine` already resumes a pipeline from any saved node, freezes upstream outputs, and replays recorded HTTP for determinism. The one thing missing was the ability to **change** the state before resuming — `replay()` deserialized `step.input_state` verbatim.

This PR adds that layer. A developer who spots a bad value at step 7 can now correct it and resume the remaining steps, instead of editing code and re-running all ten.

```bash
argus replay RUN_ID node_7 --set status=OK            # correct a value
argus replay RUN_ID node_7 --delete stale_field       # reproduce a dropped field
argus replay RUN_ID node_7 --patch fix.json           # a full patch document
argus replay RUN_ID node_7 --set status=OK --dry-run  # preview, run nothing
```

## Where this sits

Phase 1 of a step-by-step roadmap. Each step is a separate commit, small enough to review on its own.

| Step | Scope | Status |
|---|---|---|
| 1.1 | `state_patch.py` — pure patch layer. Touches no existing module | in this PR |
| 1.2 | `ReplayEngine` wiring + `RunRecord.state_patch` provenance | in this PR |
| 1.3 | `argus replay --set/--delete/--patch/--dry-run` + preview | in this PR |
| 1.4 | `/api/replay` accepts a patch (backend only) | in this PR |
| 1.5 | `/api/replay/preview` — dry run over HTTP | tracked follow-up |
| 1.6 | Frontend JSON editor in `ReplayControls.tsx` | tracked follow-up |
| 1.7 | Persist the run when a replay crashes (see below) | tracked follow-up |

Steps 1.5–1.7 are **tracked follow-ups, not omissions** — they are deliberately separate so this PR stays reviewable.

## Why this feature first

Patching a recorded state is the primitive underneath several other planned directions — injecting a malicious tool output for red-team testing, holding a field fixed to assert a downstream contract, and comparing trajectories under a changed input are all "run the graph from a mutated state." Building it once makes those additive rather than each inventing their own mechanism.

## Design decisions worth reviewing

- **Patch before deserialize.** The patch applies to the raw recorded dict, then `safe_deserialize` runs. Typed states (pydantic / dataclass / TypedDict) are therefore validated against the patched values rather than bypassing validation.
- **Fixed op order.** `delete` → `set` → `merge`, so results never depend on JSON key ordering. A path targeted by two op groups is rejected rather than silently resolved.
- **Strict by default.** A mistyped path raises with a `difflib` "did you mean" hint instead of silently adding a field; creating new keys requires `--create-missing`. In a debugging tool, a silently absorbed typo is worse than an error.
- **Path syntax matches existing signals.** Both `items[0].name` and `items.[0].name` parse identically, so the `field_path` ARGUS already reports on a failing signal pastes straight into a patch.
- **Provenance.** `RunRecord.state_patch` records the patch and round-trips through storage, so a patched replay explains its own divergence from its parent. Unpatched replays record `None`.
- **Fail before side effects.** A malformed patch raises before any node function is imported or executed — asserted by a test that checks no node ran. The same holds at the API layer: validation happens in the request handler, so a bad patch returns `400` with the hint rather than being accepted as a job and failing opaquely in a worker thread.
- **No new dependencies.**

## Testing

86 new tests (66 unit on the patch layer + CLI construction, 20 end-to-end on patched replay — 8 of which stand up the real dashboard handler on an ephemeral port and drive it over HTTP).

Coverage includes the patched trajectory, upstream freezing, provenance round-trip, immutability of the caller's state, and every rejection path.

- Full suite: **264 passed, 3 skipped** (baseline was 178 passed, 3 skipped)
- `ruff check .` — clean
- `mypy`: **164 errors before and after** — no new ones; `state_patch.py` itself is strict-clean
- Verified manually end to end through the `argus replay` CLI: preview, real patched replay, persistence, and each error path

Note: the `CI` workflow triggers only on PRs targeting `main`/`master`, so it does not run against this `dev`-targeted PR at all. The commands it would run (`ruff check .` and `pytest --tb=short`) were run locally and pass. Worth deciding separately whether `dev` should be added to that trigger list — nothing currently validates PRs into the active development branch.

## Noted, not changed — step 1.7

A node crash during replay propagates out of `_replay_direct` before `session.finalize()` is reached, so **no replay record is persisted for a crashing replay**. This is pre-existing behaviour, unrelated to patching.

It matters because "delete a field to reproduce the crash" is exactly the workflow that hits it — the reproduction currently surfaces as a raised exception with no saved forensic record. The fix is small (wrap the execution loop so finalize runs on the crash path) but it changes existing replay semantics, so it gets its own step and its own tests rather than widening this PR. The delete-op test documents current behaviour rather than asserting a record exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,33 @@ argus diff <rerun-id>                 # compare vs original
 
 External API calls (OpenAI, etc.) are recorded by default — replays are free and deterministic.
 
+### Time-Travel: edit the state, then resume
+
+Spotted the bad value? Fix it in the saved state and resume from there — no code change, no re-running the steps that already worked:
+
+```bash
+argus replay <run-id> node_7 --set status=OK          # correct a value
+argus replay <run-id> node_7 --delete stale_field     # reproduce a dropped field
+argus replay <run-id> node_7 --patch fix.json         # a full patch document
+argus replay <run-id> node_7 --set status=OK --dry-run  # preview, run nothing
+```
+
+Upstream nodes stay frozen, so only the resumed trajectory changes. Paths are dotted with list
+indices — `items[0].name` — and match the `field_path` ARGUS reports on a failing signal, so you
+can paste one straight in. A patch file takes the same three ops:
+
+```json
+{
+  "delete": ["broken_field"],
+  "set":    {"query": "fixed query", "meta.retries": 0},
+  "merge":  {"config": {"temperature": 0}}
+}
+```
+
+Patches are strict by default: a mistyped path errors with a "did you mean" hint instead of
+silently adding a field (use `--create-missing` to add new keys). Every patched replay records
+the patch it ran with, so the run explains its own divergence from the original.
+
 ---
 
 ## Semantic Judge
--- a/src/argus/cli/cmd_open_ui.py
+++ b/src/argus/cli/cmd_open_ui.py
@@ -11,6 +11,7 @@ import uuid
 import webbrowser
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from pathlib import Path
+from typing import Any
 from urllib.parse import parse_qs, unquote, urlparse
 
 from rich.console import Console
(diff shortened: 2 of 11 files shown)


abhisheksharma001:

abhisheksharma001:
**Review** — read the full diff (state_patch core, replay wiring, CLI, UI endpoints, tests). No blocking findings.

Design calls that are right:

- `_patch_input` patches the **raw recorded dict before deserialization** — typed-state validation still applies to patched values instead of being bypassed. Correct order of operations.
- Fixed op order (delete → set → merge) + `_check_conflicts` rejecting multi-op paths by parsed segments (`a[0]` ≡ `a.[0]`) — deterministic regardless of JSON key order.
- `create_missing=False` default: typos error with did-you-mean hints instead of silently adding fields.
- Provenance: patch persisted on RunRecord + storage round-trip (`storage.py` L282).
- `--set` JSON-parse-with-string-fallback + quote escape hatch documented.

76 new tests across `test_state_patch.py` (49) / `test_replay_patch.py` (27).

Minor notes, none blocking:

- `cmd_replay.py` `build_patch`: 🔵 nit — file's `set` op combined with `--set` flags overrides same-path entries silently; docstring documents it, but a one-line stdout note when an override actually happens would help debugging patched replays.
- `_write`: 🔵 nit — negative list indices accepted (`-1`) by the range check; worth one line in the README table since "patches edit existing positions" reads as 0-indexed-only.
- `_parse_set_value`: 🟡 risk — only catches `ValueError`; `json.loads("NaN")` succeeds and yields float nan, which then fails typed-state validation downstream with a confusing message. Consider rejecting non-finite floats at parse time.

CI green on my run; the two `TestCLI` e2e failures elsewhere are pre-existing on clean dev (bare `python3 -m argus.cli.main` unavailable) — unrelated.

PR #30, opened 2026-08-19
docs: roadmap backlog (good-to-have-dev) + "How We Work" process rule

abhisheksharma001:
## What this is

Documentation only. No source file is touched.

Two related things: a set of decision records for parked feature ideas, and the process convention that governs how we work through them.

## 1. `good-to-have-dev/` — decision records

A long brainstorm produced a list of "ARGUS should build X" proposals. Checking each against `src/argus/` turned up two things worth writing down:

1. **A surprising number are already built.** The proposal to "build a Root-Cause Attribution Engine" describes `correlator.py`, which has shipped for a while. "Resume execution from step N" is `ReplayEngine.replay()`. "Behavioral trajectory diffing" is `cmd_diff.py`.
2. **Several of the rest are much smaller than proposed**, because the groundwork exists. The live loop guard needs an embedding comparison, and `embedding_store.py` already provides cached `text-embedding-3-small` similarity.

Every doc follows the same shape:

> the idea, stated fairly → **what already exists** (with file paths) → **what's genuinely missing** → rough size → why it's parked

The third section is the point. The gap between "build a blame attribution engine" and "add embedding drift scoring to `correlator.py`" is a quarter versus an afternoon, and the original proposals couldn't see that distinction because they were written from the pitch rather than the code.

These map onto a phased roadmap, so the folder is a **backlog with an order**, not a loose idea dump:

| Doc | Roadmap phase | Verdict |
|---|---|---|
| `regression-contracts.md` | Phase 2 | **Build next.** Needs a golden-trace tag first, which Phase 6 also depends on |
| `graph-hazard-linter.md` | Phase 3 | **Cheapest standalone win.** Pure graph algorithms over `patcher.extract_edge_map()`. No LLM, no new deps |
| `live-loop-guard.md` | Phase 4 | Half exists — `loop_analyzer.py` is post-hoc. Argues the live version **must default to warn, not kill** |
| `flaky-tool-heatmap.md` | Phase 5 | Data already captured by `http_recorder.py` and discarded at end of run. Needs aggregation, not instrumentation |
| `smoke-test-runner.md` | Phase 6 | Largest; gateway to drift metrics and injection testing |
| `indirect-injection-testing.md` | Phase 7 | Highest *value*, lowest priority. `session.frozen_outputs` is already a working injection primitive |

### Argued against

`dropped-ideas.md` covers two proposals recommended **not** to build:

- **Rebuilding a blame attribution engine.** `correlator.py` already ships it — `DegradationOrigin` with confidence scores, `PropagationChain`, `TimelineEvent`, plus LLM augmentation in `llm_correlator.py`. The doc maps each proposed capability to where it already lives. The one genuine upgrade is embedding-based drift scoring per transition, logged as an enhancement to that module — specifically so nobody builds a second correlator next to the first.
- **RAG chunk-to-question generation.** A data-generation product, not a tracing one. No shared data model, different buyer, competes with Ragas on their home ground. The "adversarial humanization" insight behind it is genuinely sharp — it just belongs in someone else's tool. The doc suggests the ARGUS-shaped alternative: mutate *recorded* states, which `state_patch.py` (PR #29) now does.

`README.md` also carries an **"already shipped — do not re-propose"** table covering replay-from-node, state patching, blame attribution, loop analysis, behavioral diffing, HTTP determinism, custom validators, and full state capture — each with the module it lives in. Plus how ideas leave the folder: when one is built, its doc is deleted in the implementing PR.

## 2. `CONTRIBUTING.md` — "How We Work"

A short section recording that work is planned as numbered steps and executed one step at a time, stopping after each so direction can be corrected.

It's written down because the opposite happened here: a plan of three parts was agreed, and all three shipped in one pass along with two PRs, leaving no point to adjust course. The section calls out AI coding assistants specifically, since they will otherwise complete an entire roadmap in a single pass.

**Placement note:** this went in `CONTRIBUTING.md` rather than a `CLAUDE.md`, because `.gitignore` deliberately treats `CLAUDE.md` as a local settings file alongside `.claude/`. Committing one would have reversed that choice.

## Naming

The folder is `good-to-have-dev` rather than the literally requested `good to have dev` — spaces in a repo path break shell globs and tooling. Trivial to rename if the literal spelling is preferred.

## Verification

Docs only. `ruff check .` clean and the suite passes unchanged (178 passed / 3 skipped on this branch, which is cut from `dev` and deliberately contains none of the state-patching code from #29).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -122,6 +122,25 @@ For bug reports, include:
 - A minimal run output dict that reproduces the issue
 - What ARGUS reported vs what you expected
 
+## How We Work
+
+**Break work into small, trackable steps and do them one at a time.**
+
+Plan the whole thing first as numbered steps. Then execute one step at a time, stopping after each
+so the direction can be corrected before more work is built on top of it.
+
+- Don't chain steps together
+- Don't build a whole feature because the plan for it was agreed — **agreeing a plan is not
+  agreement to execute all of it at once**
+- Each step should be small enough to review in one sitting and land as its own commit
+- If a step turns out bigger than it looked, stop and re-split it rather than pushing through
+
+The point is that work stays steerable. A finished feature that went the wrong way three steps
+back costs more than the few check-ins it would have taken to catch it.
+
+**This applies to AI coding assistants too** — arguably more, since they will happily complete an
+entire roadmap in one pass. If you're using one on this repo, hold it to the same rule.
+
 ## Pull Requests
 
 - Keep PRs focused — one fix or feature per PR
--- a/good-to-have-dev/README.md
+++ b/good-to-have-dev/README.md
@@ -1,0 +1,53 @@
+# Good to have — dev
+
+Parked feature ideas for ARGUS, with an honest assessment of each.
+
+This folder exists because a long brainstorm produced a list of "ARGUS should build X" ideas, and
+a surprising number of them turned out to be **already built**. Rather than let that analysis
+evaporate into a chat log and get re-litigated in three months, each idea gets a short decision
+record here.
+
+## What each doc contains
+
+1. **The idea** — what was proposed, stated fairly
+2. **What already exists** — the specific modules in `src/argus/` that already do part of it
+3. **What's genuinely missing** — the real delta
+4. **Rough size** — honest effort estimate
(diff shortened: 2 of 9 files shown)


abhisheksharma001:

PR #32, opened 2026-08-21
Wave E: retry/cycle false positives, validator severity, replay verdict, confidence explanation

VaradDurge:
## Summary

Wave E quality work, each ticket on its own branch, merged into `dev` after per-branch tests, then the full suite on `dev`.

- **VAR-96** (includes VAR-97): superseded retry iterations no longer taint `argus check` or correlation; generate→validate→retry cycles no longer flag TC-002 ordering anomalies
- **VAR-98**: custom validators may return `(ok, message, "warning")` so soft checks do not fail the node or CI gate; `(ok, message)` still blocks
- **VAR-94**: `argus replay` / `argus show` print whether the original failing node is **RESOLVED** or **NOT RESOLVED**
- **VAR-95**: `argus show` Correlation panel includes a **Why this score?** breakdown of origin confidence

## Test plan

- [x] Per-branch tests for each VAR
- [x] Full `pytest tests/` on `dev` (677 passed; `test_cmd_init` needs a writable `.cursor` dir — passes outside sandbox)
- [ ] CI green on this PR (3.9 / 3.11 / 3.12)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ARGUS_STABILITY_REVIEW.md
+++ b/ARGUS_STABILITY_REVIEW.md
@@ -75,6 +75,45 @@ stable for fleet-scale use.
 10. **Docs**: document the JSON schema (`schema_version` exists — publish what it covers),
     the status vocabulary, and validator/anomaly field shapes.
 
+## Detection recall — fault-injection matrix (the number that matters most)
+
+Tolerating 50 good agents says little; catching broken ones is the product. We injected
+10 failure classes (one per defect type, all offline, all exiting 0 except the crash
+case) and scored ARGUS 0.10.1 heuristic-only mode. Regenerate: `.venv/bin/python
+faults/_run_matrix.py` → `faults/FAULT_MATRIX.md`.
+
+**Recall: 4/10 caught** (overall_status != clean).
+
+| caught | missed |
+|---|---|
+| dropped field (BA-006), crash root-cause (`crashed` + correct chain), placeholder output (PH-007/RF-005), validator breach | empty `{}` node update, swallowed type contract violation, loop stall, wasted retries, degraded/truncated text, implausibly fast LLM call |
+
+Root causes of the six misses, verified against installed source:
+
+1. **No-op node updates are invisible.** A node returning `{}` passes; nothing compares
+   "fields this node's consumers need" vs "fields it set". Needs consumer-aware schema
+   expectations.
+2. **Swallowed exceptions look clean.** If application code catches its own error and
+   falls back, every step reports pass. Consider flagging suspiciously broad fallbacks
+   (constant outputs after exception-shaped control flow).
+3. **Loop stall / wasted-retry analysis requires the hosted LLM proxy.**
+   `loop_analyzer.py` errors with "Not logged in" without `argus login`, leaving
+   `loop_analyses: []`. The README markets these as features; offline they silently
+   no-op. Either ship a local heuristic fallback or say plainly they're cloud-only.
+4. **Warnings don't escalate.** Degraded text earns only severity=warning
+   (`shallow_output`), step stays `pass`, run stays `clean`. Define when warnings should
+   aggregate into a failure (or expose a strictness knob).
+5. **Latency checks are unreachable by default.** `suspiciously_fast` requires
+   `min_expected_ms`; `ArgusConfig` has no default and `ArgusWatcher.__init__` doesn't
+   accept it as a kwarg — the check cannot fire on default config at all today.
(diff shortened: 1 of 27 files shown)


VaradDurge:
## Review — Wave E (VAR-94/95/96/97/98)

Reviewed the diff and exercised every ticket end-to-end with real LangGraph pipelines (running the CLI the way a user would), plus the full suite. **LGTM to merge** — two non-blocking notes at the bottom.

### Automated checks
| Check | Result |
|---|---|
| `pytest tests/` (local, py3.9) | **683 passed, 3 skipped, 2 xfailed** (5m16s) |
| `ruff check src/` | clean |
| CI — test 3.9 / 3.11 / 3.12 + CLA | all green |

`test_cmd_init` (flagged in the PR body as needing a writable `.cursor`) passed here — no sandbox.

### Behavior verified with live pipelines
Built two real `StateGraph` pipelines and drove them through `ArgusWatcher` → `argus show` / `argus check` / `argus replay`:

**VAR-96 / VAR-97 — retry/cycle false positives.** A `generate → validate → (retry) → finalize` graph where the first draft is a placeholder and the retry passes:
- earlier attempts marked `retried`, `overall_status=clean`, `argus check <id>` → **exit 0** ✅
- `analyze_tool_chains` raises **no TC-002 ordering anomaly** for the `generate↔validate` cycle ✅

**VAR-98 — validator severity.**
- 3-tuple `(ok, msg, "warning")` → `is_blocking=False`, node not failed, run stays clean ✅
- 2-tuple `(ok, msg)` on a real placeholder → `semantic_fail` + `argus check` fails ✅

**VAR-95 — confidence explanation.** `argus show` renders it for real:
```
│   Origin:   retrieve  (step 0)  confidence: 100%
│   Why this score:
│     tool_failure (critical)  +3.00
│     tool_failure (warning)  +1.50
```

**VAR-94 — replay verdict.** `argus replay <id> retrieve --app mod:factory` after fixing the node:
```
  original failure  retrieve  RESOLVED
```
`compare_replay` also returns the `not resolved` verdict when the failure persists ✅

**Regression guard.** Confirmed Wave E didn't blunt detection: a tool error (`{"docs": [], "error": ...}`) → `silent_failure`, downstream placeholder → `semantic_fail`, `argus check` correctly exits non-zero. Good — the retry/severity relaxations only suppress *superseded* and *warning* signals, not genuine failures.

### Notes (non-blocking)
1. **nit —** `_INACTIVE_STATUSES = frozenset({"retried", "skipped"})` is defined twice in `correlator.py` (lines 749 and 853). Identical, harmless; worth consolidating to one module-level constant.
2. **pre-existing, out of scope —** `argus show <id>` (and `argus show run <id>`) ignores the positional and always renders the **most recent** run — the `context_settings={"allow_extra_args": True, ...}` on the `show` command routes the bare id into `ctx.args`, leaving `run_id=None` → `show_last()`. Not introduced here (`cli/main.py` isn't in this PR), but it caps VAR-95's usefulness: you can only see the "Why this score" panel via `argus show last`, not for a specific historical run. `argus check <id>` and `argus replay <id> <node>` bind their positionals fine. Suggest a small follow-up ticket.

Not merging, per request.

abhisheksharma001:
**Review** — Wave E fixes several findings from `ARGUS_STABILITY_REVIEW.md` (the `empty_output` origin fix in `inspector.py` closes fault-01 from our recall matrix; validator severity API is back-compat with 2-tuple validators ✓). One real test bug, two nits:

`tests/test_tool_chain_analyzer.py` L132-155: 🔴 bug: `test_no_ordering_anomaly_correct_order` was deleted — its body got pasted as dead code inside `test_no_ordering_anomaly_on_retry_cycle` after the first assert. The correct-order scenario now runs only if the retry-cycle assert passes first, and failures report under the wrong test name. Restore as two separate tests.

`src/argus/correlator.py` L809+L885: 🔵 nit: `_INACTIVE_STATUSES` defined twice (same value today — shadow trap tomorrow). Keep one, module top.

`src/argus/correlator.py` `_confidence_breakdown`: 🟡 risk: weights hardcoded (crit×3.0, warn×1.5, missing×0.5, anomaly 2.0/0.8) duplicate whatever the scorer uses. If scoring constants change, breakdown silently lies. Source both from one shared table.

Context for the empty_output change: our fault-injection suite (`faults/_run_matrix.py`, see review doc §Detection recall) measured this exact miss at 4/10 recall. Happy to rerun the matrix against this branch pre/post-merge if useful evidence.

Issue #26, opened 2026-08-17
Improve inspector: detect JSON-in-string fields (double-encoded JSON)


VaradDurge:
## What

A common silent failure in LLM pipelines is double-encoded JSON — a node returns \`{\"data\": \"{\\\"key\\\": \\\"value\\\"}\"}\` instead of \`{\"data\": {\"key\": \"value\"}}\`. The downstream node receives a string where it expects an object, causing a \`KeyError\` or silent data loss.

The inspector (\`src/argus/inspector.py\`) currently has 16 rules for detecting tool failures and structural issues, but none for this pattern.

## How

Add a new rule (Rule 17) to \`inspect_tool_outputs()\`:

1. Walk the output dict looking for string values
2. Try \`json.loads()\` on strings that start with \`{\` or \`[\`
3. If it parses successfully, flag it as a \`json_in_string\` tool failure with \`warning\` severity
4. Skip fields where stringified JSON is expected (e.g., fields named \`raw_response\`, \`log\`, etc.)
5. Add tests with double-encoded and normal outputs

## Reference

Look at how Rules 1–5 in \`inspect_tool_outputs()\` detect failure shapes — this follows the same pattern. The \`ToolFailure\` dataclass in \`models.py\` is the return type.

## Scope

~30 lines of detection logic + tests. One function, one file.

Sravan1011:
@VaradDurge I'll work on this

Sravan1011:
@VaradDurge @abhisheksharma001 can close this issue

VaradDurge:
Implemented in v0.10.5 (Rule 17, #34). Inspector now flags string fields that parse as JSON object/array as `json_in_string` (warning, advisory); expected-payload fields are skipped. Live on PyPI.

PR #34, opened 2026-08-22
feat: add double-encoded json detection (rule 17)

Sravan1011:
Solves #26 
Overview

**This PR implements Rule 17**: Double-Encoded JSON Detection in the ARGUS monitoring engine. It detects silent failures where a pipeline node returns a stringified JSON object or array (e.g. {"data": "{\"key\": \"value\"}"}) instead of a parsed Python dictionary or list structure. Double-encoded JSON is a common silent failure pattern that can trigger unexpected type mismatches, downstream parsing crashes, or silent data loss.

Proposed Changes
**1. Core Monitoring & Inspection**
src/argus/inspector.py
Helper function _scan_double_encoded: Added a recursive scanner (max depth 5) that checks string values to see if they start with { or [ (ignoring leading whitespace). If a string parses successfully into a dict or list using json.loads, a ToolFailure with type json_in_string and warning severity is registered.
Skipped Fields: Configured _JSON_EXPECTED_KEYS = {"raw_response", "log", "logs", "raw", "history", "raw_output", "payload"} (case-insensitive) to skip fields where stringified JSON payloads are expected.
Integration: Integrated the scanner into inspect_tool_outputs.

**2. Fix Prompt Jargon Translation**
src/argus/fix_prompt.py
Mapped "json_in_string" in both _TOOL_FAILURE_PLAIN and _TOOL_FAILURE_CONDITION to translate the internal failure enum into clear, jargon-free explanations for code generation agents:
Plain description: "the output contained a double-encoded stringified JSON object/array instead of a parsed structure"
Condition description: "the output contains a double-encoded stringified JSON object/array instead of a parsed structure"

**3. Testing**
tests/test_inspector_unit.py
Added the TestRule17JsonInString test class to verify all aspects of the new detection rule, testing:
Stringified dictionary inputs.
Stringified list inputs.
Case-insensitive field skipping on raw/payload fields.
Recursive validation on nested objects/arrays.
Safely ignoring non-JSON strings and malformed JSON strings without raising errors.
Verification Results
Automated Unit Tests
Executed the test suite inside the virtual environment:

bash
.venv/bin/pytest tests/test_inspector_unit.py
Results:

98 tests passed (including all 6 newly added test cases for Rule 17).

### DIFF
--- a/src/argus/fix_prompt.py
+++ b/src/argus/fix_prompt.py
@@ -63,7 +63,10 @@ _TOOL_FAILURE_PLAIN = {
     "empty_result": "the external call returned an empty result",
     "error_in_data": "the data that came back contained an error payload",
     "partial_failure": "the external call only partly succeeded",
-    "json_in_string": "the output contained a double-encoded stringified JSON object/array instead of a parsed structure",
+    "json_in_string": (
+        "the output contained a double-encoded stringified JSON object/array "
+        "instead of a parsed structure"
+    ),
 }
 
 # Present tense, for the "Done when" conditions — the past-tense phrasings above
@@ -74,7 +77,10 @@ _TOOL_FAILURE_CONDITION = {
     "empty_result": "the external call returns an empty result",
     "error_in_data": "the data that comes back contains an error payload",
     "partial_failure": "the external call only partly succeeds",
-    "json_in_string": "the output contains a double-encoded stringified JSON object/array instead of a parsed structure",
+    "json_in_string": (
+        "the output contains a double-encoded stringified JSON object/array "
+        "instead of a parsed structure"
+    ),
 }
 
 
--- a/src/argus/fix_prompt.py
+++ b/src/argus/fix_prompt.py
@@ -63,6 +63,7 @@ _TOOL_FAILURE_PLAIN = {
     "empty_result": "the external call returned an empty result",
     "error_in_data": "the data that came back contained an error payload",
     "partial_failure": "the external call only partly succeeded",
+    "json_in_string": "the output contained a double-encoded stringified JSON object/array instead of a parsed structure",
 }
 
 # Present tense, for the "Done when" conditions — the past-tense phrasings above
@@ -73,6 +74,7 @@ _TOOL_FAILURE_CONDITION = {
     "empty_result": "the external call returns an empty result",
     "error_in_data": "the data that comes back contains an error payload",
     "partial_failure": "the external call only partly succeeds",
+    "json_in_string": "the output contains a double-encoded stringified JSON object/array instead of a parsed structure",
 }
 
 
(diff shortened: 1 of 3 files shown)


Sravan1011:

VaradDurge:
yo @Sravan1011 this is clean, ngl 🔥 double-encoded JSON is such a sneaky silent failure so good catch adding it. also love that you kept it at `warning` sev + the skip-list, so it's not out here failing everyone's CI gate for no reason. tests are solid too.

one thing's gonna block the merge tho, plus a couple lowkey optional notes:

**gotta fix this — CI lint (`ruff check .`) will go red**

3 lines are over the 99-char limit (E501):

- `src/argus/fix_prompt.py:66` + `:77` — the new `json_in_string` strings
- `src/argus/inspector.py:416` — the `_JSON_EXPECTED_KEYS` frozenset

CI hasn't caught it yet (i think the workflow needs a maintainer to approve it for first-time contributors), but it'll fail the sec it runs. easy fix, just wrap em:

```python
_JSON_EXPECTED_KEYS = frozenset({
    "raw_response", "log", "logs", "raw", "history", "raw_output", "payload",
})
```

run `ruff check src/` locally and it should be green after.

**optional, not deal breakers fr**

- `_scan_double_encoded` basically re-walks the whole output tree even tho `_scan_payload_for_tool_failures` already does that same recursion right above it. totally fine for now, but eventually folding it into that walk saves a second pass per node.
- tiny edge case: when a key's in the skip-list you `continue`, so nested stuff *inside* `payload` etc never gets scanned either. prob fine since those hold raw blobs anyway, just flagging it 👀

fix the lint and this is good to go on my end. W contribution 🙌

abhisheksharma001:
**Review** — feature is sound: recursion depth-capped, malformed-JSON safe, allowlist case-handled, warning severity keeps false-positive blast radius small. Tests cover dict/list/nested/allowlist/malformed.

CI lint FAILURE is almost certainly formatting: `src/argus/inspector.py` — only one blank line between the `_JSON_EXPECTED_KEYS` block and `def _scan_double_encoded` (PEP8 wants two), and `tests/test_inspector_unit.py` L155 — no blank lines between `test_malformed_json_ignored` and the `# ── Strict mode ──` section. Run the repo linter locally; both are mechanical fixes.

🟡 risk, not blocking: hardcoded `_JSON_EXPECTED_KEYS` allowlist won't cover fields that legitimately hold JSON strings — `*_json`, `body`, `response`, ES/$-style `query` payloads all start with `{`. Warning severity contains it for now; if noise shows up in real runs, consider a per-project suppression config (same mechanism the stability review asks for signature tuning).

🔵 nit: `evidence=f"...{stripped[:80]!r}"` truncates without ellipsis marker — cosmetic.

PR #36, opened 2026-08-24
Fix #31: blame empty-{} origin, persist each invoke, flag no-op nodes

### LINKED ISSUE #31: Root cause blames crash site; empty {} origin stays pass; only first invoke is persisted
## Summary

Three related failures on the headline path: `search` returns `{}` → a later node (`summarize`) raises `KeyError`. ARGUS records the crash but **blames the crash site**, **does not mark the origin as a failure**, and **only persists the first `invoke()`** on a given attached watcher.

This undercuts the README claims:

- Crash root cause: *"Traces `KeyError` at node 5 back to the upstream node that actually dropped the field"*
- Attach persistence: *"written when the outermost call returns"* / *"run is persisted automatically"*

Found in local testing of a simple retrieve → summarize graph.

---

## 1. Root cause points at the symptom

**Expected:** origin is `search` (empty `{}`). `argus fix` should target `search`.

**Actual:**

- Chain line shows `summarize → search`
- Correlation box: **Origin: `summarize`, confidence 100%**
- `argus fix` writes a prompt to patch **`summarize`**

Following that prompt adds a `.get()` guard at the crash site and **leaves the bug** in `search`.

## 2. Only the first `invoke` per watcher is persisted

**Expected:** each outermost `app.invoke()` writes its own run (README: persist when the outermost call returns).

**Actual:** one `ArgusWatcher().attach(graph)`, then 3× `app.invoke()` → **1 run in storage** (the first).

A script that ran **clean then failing** stored only the clean run and printed nothing. The failure was invisible until the failing invoke was re-run alone.

Likely cause: session sets `_completed` after the first finalize; later `finalize()` is a no-op and no new `run_id` is opened.

## 3. `search` returning `{}` is never itself a failure

**Expected:** empty/dropped output on the origin is a silent failure (README demo: node returns `{}` or drops a required field).

**Actual:** `search` stays **`pass`** or pass-with-warnings (green). Only the downstream crash is red. That is why (1) can treat `summarize` as origin.

---

## Reproduction (minimal)

LangGraph: `search` → `summarize`.

- `search` returns `{}` (or omits the field `summarize` reads).
- `summarize` does `state["documents"]` (or similar) and raises `KeyError`.

```python
from argus import ArgusWatcher

watcher = ArgusWatcher()
app = watcher.attach(graph)

app.invoke(clean_input)     # recorded
app.invoke(failing_input)   # not recorded; no [argus] print
```

Then: `argus show last` / Correlation / `argus fix <id>` on a run that *did* crash.

## Environment

- `argus-agents` from current `master`
- Heuristic-only is enough (no LLM judge required)

## Done when

- [ ] Origin node that returned `{}` / dropped the required field is not a clean `pass`
- [ ] Correlation + `root_cause_chain` + `argus fix` target that origin, not the `KeyError` site
- [ ] Each outermost `invoke()` / `ainvoke()` / `stream()` on an attached app persists its own run (or documented otherwise, with a new session per call)
- [ ] README claims match the behavior

## Notes

These three are one story: empty origin stays green → blame/fix follow the crash → a second invoke never gets a run. Prefer one PR (or a stacked series) over three disconnected patches.

If you want this issue, comment below and a maintainer will assign you. Work from a **fork** — write access on this repo is not required.

abhinavships:
Closes #31.

### DIFF
--- a/src/argus/inspector.py
+++ b/src/argus/inspector.py
@@ -841,6 +841,27 @@ def inspect_transition(
         else None
     )
     tool_failures = _tool_result.tool_failures if _tool_result is not None else []
+
+    # Silent no-op: the node ran but returned an empty state update ({}) — no
+    # keys at all. With successors waiting downstream it has produced nothing
+    # for them to consume — the canonical silent failure the ORIGIN node commits
+    # (README: "node returns {} or drops a required field"). Without this the
+    # drop is invisible and blame falls on the downstream crash site instead.
+    # Only literal {} is flagged here: a dict WITH keys (even empty-valued ones,
+    # e.g. {"vulnerabilities": []}) is a real state contribution and is judged
+    # by the per-field rules above. Conditional/router nodes reach here with
+    # successor_fns=[] (see ArgusSession._get_successor_fns) and are exempt.
+    if successor_fns and output_dict is not None and not output_dict:
+        tool_failures = [
+            *tool_failures,
+            ToolFailure(
+                failure_type="empty_output",
+                field_name="_output",
+                severity="critical",
+                evidence="node returned an empty state update — no fields produced",
+            ),
+        ]
+
     has_tool_failure = any(tf.severity == "critical" for tf in tool_failures)
 
     if output_dict is None:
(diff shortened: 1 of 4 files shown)


abhinavships:

PR #39, opened 2026-08-25
fix(show): nonzero exit on unresolvable run ids + clean ambiguous-prefix errors

### LINKED ISSUE #33: argus show <id> ignores the positional id and always shows the most recent run
**Priority: High** (correctness bug in a core inspect command — silent wrong output; workaround exists so not Urgent)

## Summary
`argus show <id>` ignores the positional run id and always renders the **most recent** run. It fails silently — no error — so a user inspecting run X is shown run Y instead.

## Repro
With ≥2 stored runs:
```
argus show <full-run-id>      # shows last_run_id(), not <full-run-id>
argus show run <run-id>       # back-compat form — also shows the last run
argus show last               # correct (this is the only working path today)
```

## Cause
`cli/main.py:264` decorates the `show` command with
`context_settings={"allow_extra_args": True, "allow_interspersed_args": True}`.
The bare id is routed into `ctx.args` instead of binding to the declared
`run_id` argument, so `run_id` stays `None` → the `run_id is None` branch calls
`show_last()`.

`load_run(<id>)` / `_resolve_run_path` are correct — verified they return the
right record directly. The defect is purely in CLI arg routing on the `show`
command.

## Impact
- Silently shows the wrong run — undermines trust in the inspect surface.
- Caps **VAR-95** (confidence "Why this score" breakdown): only viewable via
  `argus show last`, never for a specific historical run.
- `argus check <id>` and `argus replay <id> <node>` are **unaffected** — they
  bind positionals normally (no `allow_extra_args`).

## Suggested fix
- Remove `allow_extra_args` from the `show` command; preserve the
  `argus show run <id>` back-compat via an explicit fallback that reads the id.
- Add a regression test: with two runs, `show <first-id>` renders the first id.

## Discovered
While reviewing #32 (Wave E). Pre-existing — not introduced by that PR.

abhisheksharma001:
## Problem

Three error paths in `argus show` misbehave for anyone scripting the CLI:

1. **Unknown id exits 0.** `argus show <bad-id>` prints `Error: …` then returns exit code 0 — success and failure are indistinguishable.
2. **Ambiguous prefix dumps a traceback.** A prefix matching multiple runs raises `ValueError` that only the `--json` family half-handles; the human path renders a full rich traceback panel.
3. **Latent crash on empty store.** `argus show last --json` with no runs calls `Console.print(..., err=True)` — not a real rich kwarg — so the intended friendly `No runs found.` message becomes a `TypeError` traceback.

(`argus check` / `argus fix` already exit 1 correctly on bad ids — this brings `show` in line.)

## Fix

- New `_show_run_strict()` helper resolves the id via `load_run()` first; on `FileNotFoundError`/`ValueError` prints one clean error line and exits 1. Used by both `show <id>` and the back-compat `show run <id>` form.
- `--json` path additionally catches `ValueError`.
- Dropped the invalid `err=True` kwargs (3 sites).

## Tests

`tests/test_cmd_show_errors.py` — 7 cases: full id / unique prefix / back-compat `run <id>` happy paths render the *requested* run (not newest); unknown id → exit 1; ambiguous prefix → exit 1 with message and no `Traceback` in output; both json error paths covered. All red before the fix, green after. Full suite: no regressions (only the two pre-existing `test_e2e_langgraph.py::TestCLI` env failures remain, which shell out to bare `python3 -m argus.cli.main` and fail identically on clean `dev` without an installed argus).

## Notes

- Verified empirically that #33's core claim (positional id ignored) does **not** reproduce on PyPI 0.10.1 nor current `dev` — all show forms resolve correctly. This PR hardens the adjacent error paths that issue surfaced; happy to close #33 alongside if maintainers confirm.

### DIFF
--- a/src/argus/cli/main.py
+++ b/src/argus/cli/main.py
@@ -281,12 +281,15 @@ def cmd_show(
         if run_id == "run" and ctx.args:
             target_id = ctx.args[0]
         if target_id is None:
-            _console.print("[red]Error:[/red] No runs found.", err=True)
+            _console.print("[red]Error:[/red] No runs found.")
             raise typer.Exit(1)
         try:
             print(load_run_text(target_id))
         except FileNotFoundError as e:
-            _console.print(f"[red]Error:[/red] {e}", err=True)
+            _console.print(f"[red]Error:[/red] {e}")
+            raise typer.Exit(1)
+        except ValueError as e:
+            _console.print(f"[red]Error:[/red] {e}")
             raise typer.Exit(1)
         return
     if run_id is None or run_id == "last":
@@ -295,11 +298,32 @@ def cmd_show(
         # Backward compat: 'argus show run <id>' still works
         actual_id = ctx.args[0] if ctx.args else None
         if actual_id:
-            show_run(actual_id)
+            _show_run_strict(actual_id)
         else:
             show_last()
     else:
-        show_run(run_id)
+        _show_run_strict(run_id)
+
+
+def _show_run_strict(run_id: str) -> None:
+    """Render a specific run, exiting nonzero when the id can't be resolved.
+
+    ``show_run()`` prints errors and returns so the TUI stays friendly, but a
+    caller scripting ``argus show <id>`` then can't tell success from failure —
+    and an ambiguous prefix escaped as a raw traceback. Resolve first; on any
+    resolution failure print the reason and exit 1.
+    """
+    from argus.storage import load_run
+
+    try:
+        load_run(run_id)
+    except FileNotFoundError as e:
+        _console.print(f"[red]Error:[/red] {e}")
+        raise typer.Exit(1)
+    except ValueError as e:
+        _console.print(f"[red]Error:[/red] {e}")
+        raise typer.Exit(1)
+    show_run(run_id)
 
 
 @app.command("check")
--- a/tests/test_cmd_show_errors.py
+++ b/tests/test_cmd_show_errors.py
@@ -1,0 +1,88 @@
+"""``argus show <id>`` must fail loudly when the id can't be resolved.
+
+Companion to the #33 family: positional ids now resolve correctly, but the
+error paths still exited 0 (unknown id) or dumped a raw traceback (ambiguous
+prefix) — either way a script can't tell success from failure.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timedelta, timezone
+
+from conftest import make_event, make_run_record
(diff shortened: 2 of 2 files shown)


VaradDurge:
## Review — routing bug: this doesn't fix #33, and the tests fail on a clean checkout

Thanks for this — the `_show_run_strict` helper and the `err=True` cleanup are sound. But I can't merge it yet: **`argus show <positional-id>` still ignores the id and renders the last run at exit 0**, so #33 is not actually fixed, and 5 of the 7 new tests fail on a fresh checkout of this branch. CI only ran the CLA check here — the test job never ran, so "green after" wasn't verified by the pipeline.

### Reproduction
On `fix/show-unresolvable-id-exit-codes`:
```
$ pytest tests/test_cmd_show_errors.py -q
5 failed, 2 passed
```
Tracing `show <unknown-id>` end to end:
- `load_run("20990101-…zzzz9999")` correctly raises `FileNotFoundError` — the helper logic is fine.
- But the CLI routes to **`show_last()`**, not `_show_run_strict()`. Instrumented: `ROUTING: {'show_last': True}`, and `load_run` is called with the *newest* run's id, not the requested one.

### Root cause
`@app.command("show", context_settings={"allow_extra_args": True, "allow_interspersed_args": True})` (main.py:264, pre-existing, untouched here). With those settings Click doesn't bind the positional `run_id` — it lands in `ctx.args`, so `run_id is None` → `show_last()`. The new `_show_run_strict(run_id)` lives in the `else` branch that **never runs for the positional path**, so it's dead code for exactly the case #33 is about. It only helps the `show run <id>` form (which reads `ctx.args[0]`) and `--json`.

### Why it looked green
The passing tests save a single run or none, so `show_last()` coincidentally renders the requested/only run (or the empty-store path exits 1). The 5 failures are the tests that save **two** runs or expect a real error — `show_last()` renders the *newer* run at exit 0. A Typer/Click version difference likely explains why positional binding behaved differently on your machine; all the more reason to gate this on CI.

### Suggested fix (root cause)
In `cmd_show`, when `run_id` is None but `ctx.args` has a value, treat `ctx.args[0]` as the id (mirroring the `show run <id>` path) — or drop `allow_interspersed_args`/`allow_extra_args` so the positional binds. Then `_show_run_strict` is reached for `show <id>`, and please re-run in CI with a fixture that saves ≥2 runs so the test proves the *requested* id (not last) is rendered.

Happy to pair on the routing fix if useful.

PR #40, opened 2026-08-25
fix(key): nonzero exit on failed key set/use

abhisheksharma001:
## Problem

`argus key use <unknown>` prints `Unknown provider '…'` and **exits 0**. Same for:

- `key use <provider>` when that provider has no saved key,
- `key set --provider <unknown>`.

Anything scripting BYOK setup gates on exit codes and currently reads failed operations as success.

## Fix

- `key_set()` / `key_use()` return `bool` (False on every failure branch — unchanged messages/output otherwise).
- Typer wrappers raise `typer.Exit(1)` on False.
- No behavior change on success paths.

## Tests

`tests/test_cmd_key_errors.py`: unknown provider (use + set) → exit 1; unsaved-key switch → exit 1; saved-key switch → exit 0 and activates exactly the requested provider (`user_config` calls monkeypatched — no writes to real `~/.argus/config.json`). Full suite: only the two pre-existing env-dependent `TestCLI` e2e failures remain (fail identically on clean `dev`).

Companion to #39 — same error-path-exit-code disease, different command family.

### DIFF
--- a/src/argus/cli/cmd_key.py
+++ b/src/argus/cli/cmd_key.py
@@ -32,17 +32,20 @@ def _validate_provider(provider: str) -> bool:
     return True
 
 
-def key_set(value: str | None = None, provider: str = "openai") -> None:
-    """Save a provider key locally and make it the active provider."""
+def key_set(value: str | None = None, provider: str = "openai") -> bool:
+    """Save a provider key locally and make it the active provider.
+
+    Returns False when the operation failed (unknown provider, empty key).
+    """
     if not _validate_provider(provider):
-        return
+        return False
     if not value:
         import getpass
 
         value = getpass.getpass(f"{provider} API key (input hidden): ").strip()
     if not value:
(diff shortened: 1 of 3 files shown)


VaradDurge (approved):
## Approve ✅

Verified on a clean checkout of `fix/key-command-exit-codes`: `tests/test_cmd_key_errors.py` → **4 passed**.

- Clean fix: `key_set()`/`key_use()` return `bool` (False on each failure branch, messages unchanged), wrappers `raise typer.Exit(1)` on False. Success paths untouched.
- No routing trap — `key use`/`key set` bind a plain positional `provider`, so the fix actually runs (unlike the `show` positional path in #39).
- No collateral: the only callers are the two Typer wrappers this PR updates; nothing consumed the old `None` return.
- Tests assert both exit code and that exactly the requested provider is activated — not green-by-coincidence.

Merging into `dev`.

PR #43, opened 2026-08-27
feat: detect truncation and stale-context phrases in LLM output

### LINKED ISSUE #23: Add new semantic signatures for common LLM degradation patterns
## What

`src/argus/data/signatures.json` is the registry of semantic failure patterns ARGUS uses to detect placeholder/degraded LLM outputs. It currently has 72 signatures across 8 categories. Several common degradation patterns are missing:

**Suggested additions:**
- **Lazy list truncation**: "...and many more", "etc.", "and so on" when a node was expected to enumerate all items
- **Self-referential confusion**: "As mentioned above" / "As I said earlier" in a single-turn node (no prior context)
- **Markdown artifacts in structured output**: Raw `###` headers or `**bold**` markers in fields expecting plain text
- **Language mixing**: Response switches language mid-output (e.g., English prompt → partial Chinese response)
- **Token limit warnings**: "I've reached my token limit" / "Due to length constraints"

## How

1. Add new entries to the appropriate category in `src/argus/data/signatures.json`
2. Each entry needs: `id`, `category`, `pattern`, `match_strategy`, `severity`, `description`
3. Follow the existing ID scheme (e.g., `PH-xxx` for placeholder_outputs, `SP-xxx` for suspicious_phrases)
4. Add unit tests in `tests/` that verify the new signatures match expected inputs and don't false-positive on normal text

## Reference

Look at existing signatures for format examples. Match strategies available: `exact_ci`, `contains_ci`, `prefix_ci`, `regex`, `repetition`.

## Scope

Pick 3–5 patterns to start. Each is ~10 lines of JSON + a test case.

AaryanPuri:
Closes #23.

Adds SP-019..026 to signatures.json (suspicious_phrases, warning severity):
- truncation admissions ("reached my token limit", "response was truncated")
- back-references to missing context ("as mentioned above")
- trailing "and so on" / "and many more" (end-anchored regex)

Tests in tests/test_registry_unit.py. Skipped the markdown-artifact and
language-mixing ideas from the issue — too false-positive-prone in a
context-free registry.

### DIFF
--- a/src/argus/data/signatures.json
+++ b/src/argus/data/signatures.json
@@ -772,7 +772,7 @@
       "pattern": "as mentioned above",
       "match_strategy": "contains_ci",
       "severity": "warning",
-      "description": "Back-reference to absent prior context in a single-turn node — model assumes conversation history that isn't present",
+      "description": "Back-reference to earlier content — the referenced context may not exist in what this node was given",
       "source": "builtin",
       "metadata": {
         "confidence": null,
@@ -787,7 +787,7 @@
       "pattern": "as i said earlier",
       "match_strategy": "contains_ci",
       "severity": "warning",
-      "description": "Self-referential back-reference — model assumes an earlier turn that doesn't exist",
+      "description": "Self-referential back-reference to a previous turn — may assume conversation history the node never received",
       "source": "builtin",
       "metadata": {
         "confidence": null,
@@ -802,7 +802,7 @@
       "pattern": "as previously mentioned",
       "match_strategy": "contains_ci",
       "severity": "warning",
-      "description": "Back-reference to absent prior context — model assumes information it was never given",
+      "description": "Back-reference to previously supplied information — may assume context the node was never given",
       "source": "builtin",
       "metadata": {
         "confidence": null,
@@ -814,7 +814,7 @@
     {
       "id": "SP-025",
       "category": "suspicious_phrases",
-      "pattern": "\\band many more\\.?\\s*$",
+      "pattern": "\\band many more[.…]*\\s*$",
       "match_strategy": "regex",
       "severity": "warning",
       "description": "Output ends with 'and many more' — enumeration abandoned mid-list",
@@ -829,7 +829,7 @@
     {
       "id": "SP-026",
       "category": "suspicious_phrases",
-      "pattern": "\\band so on\\.?\\s*$",
+      "pattern": "\\band so on[.…]*\\s*$",
       "match_strategy": "regex",
       "severity": "warning",
       "description": "Output ends with 'and so on' — enumeration abandoned mid-list",
--- a/src/argus/data/signatures.json
+++ b/src/argus/data/signatures.json
@@ -722,6 +722,126 @@
       }
     },
     {
+      "id": "SP-019",
+      "category": "suspicious_phrases",
+      "pattern": "reached my token limit",
+      "match_strategy": "contains_ci",
+      "severity": "warning",
+      "description": "Model reports hitting its token limit — output is truncated",
+      "source": "builtin",
+      "metadata": {
+        "confidence": null,
+        "frequency": null,
+        "approval_status": "approved",
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-020",
+      "category": "suspicious_phrases",
+      "pattern": "due to length constraints",
+      "match_strategy": "contains_ci",
+      "severity": "warning",
+      "description": "Model truncated its output citing length constraints",
+      "source": "builtin",
+      "metadata": {
+        "confidence": null,
+        "frequency": null,
+        "approval_status": "approved",
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-021",
+      "category": "suspicious_phrases",
+      "pattern": "response was truncated",
+      "match_strategy": "contains_ci",
+      "severity": "warning",
+      "description": "Model states its own response was truncated — incomplete output",
+      "source": "builtin",
+      "metadata": {
+        "confidence": null,
+        "frequency": null,
+        "approval_status": "approved",
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-022",
+      "category": "suspicious_phrases",
+      "pattern": "as mentioned above",
+      "match_strategy": "contains_ci",
+      "severity": "warning",
+      "description": "Back-reference to absent prior context in a single-turn node — model assumes conversation history that isn't present",
+      "source": "builtin",
+      "metadata": {
+        "confidence": null,
+        "frequency": null,
+        "approval_status": "approved",
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-023",
+      "category": "suspicious_phrases",
+      "pattern": "as i said earlier",
+      "match_strategy": "contains_ci",
+      "severity": "warning",
+      "description": "Self-referential back-reference — model assumes an earlier turn that doesn't exist",
+      "source": "builtin",
+      "metadata": {
+        "confidence": null,
+        "frequency": null,
+        "approval_status": "approved",
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-024",
+      "category": "suspicious_phrases",
+      "pattern": "as previously mentioned",
+      "match_strategy": "contains_ci",
+      "severity": "warning",
+      "description": "Back-reference to absent prior context — model assumes information it was never given",
+      "source": "builtin",
+      "metadata": {
+        "confidence": null,
+        "frequency": null,
+        "approval_status": "approved",
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-025",
+      "category": "suspicious_phrases",
+      "pattern": "\\band many more\\.?\\s*$",
+      "match_strategy": "regex",
+      "severity": "warning",
+      "description": "Output ends with 'and many more' — enumeration abandoned mid-list",
+      "source": "builtin",
+      "metadata": {
+        "confidence": null,
+        "frequency": null,
+        "approval_status": "approved",
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-026",
+      "category": "suspicious_phrases",
+      "pattern": "\\band so on\\.?\\s*$",
+      "match_strategy": "regex",
+      "severity": "warning",
+      "description": "Output ends with 'and so on' — enumeration abandoned mid-list",
+      "source": "builtin",
+      "metadata": {
+        "confidence": null,
+        "frequency": null,
+        "approval_status": "approved",
+        "framework_specific": null
+      }
+    },
+    {
       "id": "MP-001",
       "category": "malformed_payload",
       "pattern": "^\\{\\s*\\}$",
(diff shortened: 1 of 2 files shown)


AaryanPuri:

abhisheksharma001:
Reviewed this by checking out the branch locally, running the registry tests, and probing the new signatures against the real loaded registry. Tests pass (42 passed in `test_registry_unit.py`) and CI is green. Nice scoping — the FP-guard tests and the explicit end-anchor test are more rigor than the issue asked for, and skipping the markdown-artifact and language-mixing ideas was the right call.

Two things I think are worth addressing before merge, then some smaller notes.

### 1. SP-025/026 miss the most common truncation form (trailing ellipsis)

`\.?` allows exactly one dot, so the ellipsis variants don't match. Run against the branch:

```
"Affected files: a.py, b.py, c.py, and so on..."   -> []        # no match
"Stages: init, build, and many more..."            -> []        # no match
"Stages: init, build, and so on…"                  -> []        # no match  (U+2026)
"Stages: init, build, and so on.\n"                -> SP-026    # matches
```

`and so on...` is probably the single most likely way a model abandons an enumeration, so this misses the main case. Suggested pattern:

```
\band so on[.…]*\s*$
\band many more[.…]*\s*$
```

Also, `etc.` was listed in #23 under lazy list truncation and isn't covered here — worth either adding or calling out as deliberately skipped alongside the other two skips.

### 2. Warning severity isn't advisory — a hit flips node status

In `session.py:1071-1076` (and `:800-804`), any semantic signal sets `_has_signals`, which moves the node from `pass` to `semantic_fail`. Severity never gates that decision. So a false positive isn't log noise, it's a failed node.

That matters for SP-022/023/024, which the context-free registry can't disambiguate:

```
"Section 3 covers the retry policy. As mentioned above, the totals reconcile to $4,120."
  -> SP-022   # legitimate self-reference within one long output

"As I said earlier, you need to run migrations first."
  -> SP-023   # legitimate in a multi-turn chat agent
```

The registry has no way to tell "back-reference to absent context" from "back-reference to my own earlier paragraph", or from a genuine prior turn in a conversational agent.

To be fair, this is consistent with existing design rather than a new problem — SP-002 (`I cannot`), SP-006 (`I apologize`), SP-009 (`I'm not sure`) are equally broad and already ship. SP-019/020/021 don't have this ambiguity at all (those phrases are always a defect), and SP-025/026 are end-anchored so they're low risk. My suggestion would be to land SP-019/020/021/025/026 as-is, and make an explicit call on SP-022/024 — either drop them or accept them knowingly as the first candidates for the disable list.

### Smaller notes

- **End-anchored signatures can't fire on the outputs that truncate most.** `heuristic_engine.py:10,71` skips strings over 10,000 chars, and `inspector.py:810-818` replaces strings ≥50K with a preview sentinel. So SP-025/026 only reach outputs ≤10k. Not this PR's problem, but it does cap the feature's reach — long generations are exactly where enumeration gets abandoned.
- **SP-022 and SP-024 can both fire on the same string**, producing two signals for one defect. The dedup key is `(sig_id, path)`, so they don't collapse.
- **SP-022's description says "in a single-turn node"** but the matcher has no turn awareness — the description promises a discrimination the code doesn't make.
- **`CONTRIBUTING.md:33`** asks signature contributions to "include a fixture run that demonstrates the detection". This PR has unit tests but no fixture run — either add one or the doc needs updating. Relatedly `CONTRIBUTING.md:25` still says 61 patterns; it's 80 after this PR (72 before), so it was already stale and is cheap to fix here.

For what it's worth, I checked all of SP-001..018 for overlap and there's none, no ID collisions, and the metadata shape matches the other builtins exactly. `test_na_not_newly_worsened` greps `startswith("SP-01")` but intersects against a fixed set, so SP-019 slips past it harmlessly.

abhisheksharma001 (commented):
Following up on my earlier comment — here are the three description rewords as click-to-apply suggestions, so they're not extra work for you.

Separately: the `etc.` gap from #23 is now #45, and the stale registry counts in CONTRIBUTING/CLAUDE.md are #44. Neither touches this PR's lines. The ellipsis fix (SP-025/026) and the SP-022/024 keep-or-drop call are still yours.

src/argus/data/signatures.json
```diff
@@ -721,6 +721,126 @@
...
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-022",
+      "category": "suspicious_phrases",
+      "pattern": "as mentioned above",
+      "match_strategy": "contains_ci",
+      "severity": "warning",
+      "description": "Back-reference to absent prior context in a single-turn node — model assumes conversation history that isn't present",
```
abhisheksharma001: The matcher has no turn awareness — it can't tell an unresolvable back-reference from a legitimate one, so this description promises a discrimination the code doesn't make. Suggested wording that describes what actually matched:

```suggestion
      "description": "Back-reference to earlier content — the referenced context may not exist in what this node was given",
```

src/argus/data/signatures.json
```diff
@@ -721,6 +721,126 @@
...
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-023",
+      "category": "suspicious_phrases",
+      "pattern": "as i said earlier",
+      "match_strategy": "contains_ci",
+      "severity": "warning",
+      "description": "Self-referential back-reference — model assumes an earlier turn that doesn't exist",
```
abhisheksharma001: The matcher has no turn awareness — it can't tell an unresolvable back-reference from a legitimate one, so this description promises a discrimination the code doesn't make. Suggested wording that describes what actually matched:

```suggestion
      "description": "Self-referential back-reference to a previous turn — may assume conversation history the node never received",
```

src/argus/data/signatures.json
```diff
@@ -721,6 +721,126 @@
...
+        "framework_specific": null
+      }
+    },
+    {
+      "id": "SP-024",
+      "category": "suspicious_phrases",
+      "pattern": "as previously mentioned",
+      "match_strategy": "contains_ci",
+      "severity": "warning",
+      "description": "Back-reference to absent prior context — model assumes information it was never given",
```
abhisheksharma001: The matcher has no turn awareness — it can't tell an unresolvable back-reference from a legitimate one, so this description promises a discrimination the code doesn't make. Suggested wording that describes what actually matched:

```suggestion
      "description": "Back-reference to previously supplied information — may assume context the node was never given",
```

AaryanPuri:
Pushed a fix for the ellipsis case on SP-025/026. You were right, `\.?` only ever matched one dot, so the actual "and so on..." form never fired. Should catch it now, along with the `…` character version. Also took your three suggested rewords as-is, thanks for writing those out.

On SP-022/023/024, going back and forth on this, but landing on keeping them. Yeah, they'll false-positive on things like a long single-turn answer that says "as mentioned above" about itself. But that's really a symptom of `warning` severity not actually being advisory anywhere in the pipeline yet, not something this PR is introducing. Feels like the right fix is making severity actually gate node status, and I'd rather do that properly in its own PR than rush it in here. Will file it separately.

VaradDurge (commented):
Reviewed this end to end — tests are green (registry 45/45, smoke 95/95, full suite exit 0), JSON is valid, no duplicate IDs, and the ellipsis regex fix in the second commit is a genuinely good catch. The **truncation** signatures (SP-019/020/021) and the **end-anchored enumeration** ones (SP-025/026) are specific and low-risk — happy to see those land.

One blocker on **SP-022 / SP-023 / SP-024** (`as mentioned above`, `as i said earlier`, `as previously mentioned`), and it's about the justification, not the phrases themselves.

The commit message keeps them on the grounds that:

> severity=warning doesn't currently gate node status (see session.py:1071-1076)

That isn't accurate. Severity doesn't gate **tool-failures** (`has_tool_failure` only counts `critical`, session.py:1061-1062) — but suspicious-phrase matches don't go through that path. They go into `inspection.semantic_signals` (inspector.py:847, which stores *every* signal regardless of severity), and status is gated on **presence** in that list, not severity:

- session.py:786-804 (main path) and 1070-1076 (disambiguation path): `_has_signals = bool(inspection.semantic_signals)` → with no critical failure, the node is set to `semantic_fail`.

Empirically, a completely benign sentence trips it:

```
input: {"answer": "As mentioned above, the totals reconcile."}
→ SIGNALS: [('SP-022', 'warning', 0.62)]
→ has_tool_failure(critical): False
→ _has_signals: True  →  node status = semantic_fail
```

That deterministic `semantic_fail` is recorded immediately (session.py:840-842). The LLM judge that could dismiss it only runs when `_llm_investigation_config.enabled` — so in the default heuristic-only / offline / `pytest --argus` path, nothing walks it back. Net effect: a benign "as mentioned above" anywhere in node output flips the run to not-clean and fails `argus check` / `pytest --argus` — exactly the CI-gating consequence the PR assumes can't happen.

(For contrast — this is why `json_in_string` / Rule 17 can be "advisory, never flips status": it's a `warning` tool-failure, not a registry signal, so it stays out of `semantic_signals`. There's currently no equivalent "advisory" lane for a registry signature.)

Suggested resolution:
- Land SP-019/020/021/025/026 as-is.
- Hold SP-022/023/024 until there's an actual non-gating lane for them (the disable-list idea referenced in the second commit), since as written they will fire on normal prose and gate CI.

VaradDurge:
Status update after #45 merged — two things to sort before this can land:

**1. Trivial test-file conflict (needs a rebase).**
#45 (SP-027) merged to `master` and appended its own `TestEtcLazyTruncation` class at the end of `tests/test_registry_unit.py` — the same spot this PR appends `TestDegradationPhraseSignatures`. So this branch now conflicts there. It's purely an append-point collision: **no signature-ID overlap** (this PR owns SP-019..026, #45 owns SP-027), and no logic overlap. A `git rebase origin/master` (or merge master in) and keeping both classes resolves it.

**2. The SP-022/023/024 block is contingent on the gating behavior — and #47 is the fix for exactly that.**
My earlier review blocked SP-022/023/024 because warning-severity suspicious-phrase matches currently land in `inspection.semantic_signals` and flip a node to `semantic_fail` regardless of severity (session.py:786-804), so benign prose like "as mentioned above" fails the run in the no-LLM-judge path.

PR #47 ("warning-severity semantic signals no longer fail a node on their own") targets precisely that behavior. If #47 lands and does what it says, the blocking rationale for SP-022/023/024 disappears — they'd become genuinely advisory. So the cleanest sequencing is:

- **#47 merges first** → then this PR can keep all 8 signatures as-is; or
- **#47 is dropped** → split this PR to land SP-019/020/021/025/026 now and hold SP-022/023/024.

Either way SP-019/020/021 (truncation) and SP-025/026 (end-anchored enumeration) are good to go — no dependency on #47.

Nothing to change on your side right now beyond the rebase; the SP-022/024 decision is coupled to #47's outcome. I'll follow up here once #47 is reviewed.

VaradDurge (approved):
Unblocked. #47 (merged) makes warning-severity signals advisory, so SP-022/023/024 no longer fail nodes — verified 'As mentioned above' now records as pass with the signal advisory-only. Resolved the SP-027 test/data conflict from #45 (union: SP-019..027 all present, 81 sigs, no dups). Registry tests 56/56 green on the merge.

PR #45, opened 2026-08-28
feat: detect enumeration abandoned with a trailing "etc."

abhisheksharma001:
Picks up the one lazy-truncation phrase from #23 that #43 doesn't cover. #23 lists `"...and many more"`, `"etc."`, `"and so on"` together; #43 ships the first and third and skips `etc.` without comment.

### The signature

```
(?<![/\\])\b(etc|et cetera)[.…]*\s*$
```

End-anchored like SP-025/026, so `etc.` mid-sentence never fires — only an output that *stops* at it. Two guards beyond the plain anchor:

- **negative lookbehind for `/` and `\`** — otherwise `"Config lives under /etc"` matches, and `/etc` paths are everywhere in tool output
- **`\b`** — otherwise `fetcher` and `etcher` match

`et cetera` spelled out rides along in the same alternation. Trailing ellipsis is handled (`[.…]*`), including U+2026.

### Verified against the loaded registry

```
'Stages: init, build, deploy, etc.'      -> ['SP-027']
'The /etc/passwd file was modified'      -> []
```

Tests are in `TestEtcLazyTruncation`: 5 positives (uppercase, bare `etc`, trailing `...`, `et cetera`) and 5 FP guards (mid-sentence use, two `/etc` paths, `fetcher`, `etcher`).

Full suite: 687 passed, same 28 pre-existing failures as `master` on this machine (packaging/plugin tests that need `pip install -e`, unrelated). `ruff check` clean.

### One coordination note

I numbered this **SP-027** on the assumption #43 lands first and takes SP-019..026. If #43 is dropped or renumbered, say the word and I'll renumber this to SP-019 — the id is an opaque key, nothing keys off the sequence.

Both PRs append a test class to the end of `tests/test_registry_unit.py`, so whichever merges second may need a trivial conflict resolution there. No overlap in `signatures.json` — different hunks.

https://claude.ai/code/session_01YYRgNW4XPJ72BH6uLSf6Xo

### DIFF
--- a/src/argus/data/signatures.json
+++ b/src/argus/data/signatures.json
@@ -722,6 +722,21 @@
       }
     },
     {
+      "id": "SP-027",
+      "category": "suspicious_phrases",
+      "pattern": "(?<![/\\\\])\\b(etc|et cetera)[.\u2026]*\\s*$",
+      "match_strategy": "regex",
+      "severity": "warning",
+      "description": "Output ends with 'etc.' — enumeration abandoned mid-list",
+      "source": "builtin",
+      "metadata": {
+        "confidence": null,
+        "frequency": null,
+        "approval_status": "approved",
+        "framework_specific": null
+      }
+    },
+    {
       "id": "MP-001",
       "category": "malformed_payload",
       "pattern": "^\\{\\s*\\}$",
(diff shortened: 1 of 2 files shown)


VaradDurge (approved):
Same end-anchored family as SP-025/026 (approved). Regex guards verified by adversarial probing: /etc paths, fetcher/etcher/sketcher substrings, and mid-sentence etc. all correctly excluded; bare/trailing etc. fires. 134 tests pass. No ID collision on master.

PR #47, opened 2026-08-28
fix: warning-severity semantic signals no longer fail a node on their own

### LINKED ISSUE #46: Warning-severity signature matches still flip nodes to semantic_fail
Right now severity on a signature match doesn't really change anything downstream. Whether it's `warning` or `critical`, any hit at all sets `_has_signals` and the node goes from `pass` to `semantic_fail`. See session.py:1071-1076 and :800-804.

So a `warning` match isn't advisory in practice, it fails the node the same way a `critical` one does. This came up during review on PR #43 with SP-022/023/024 (phrases like "as mentioned above"). Those are inherently ambiguous in a context-free registry, there's no way to tell a legitimate self-reference from a model assuming context it doesn't have, but they still shipped because dropping them wouldn't have actually fixed anything, just hidden the gap. Same problem already exists for SP-002 ("I cannot"), SP-006 ("I apologize"), and SP-009 ("I'm not sure"), so it's not new, just more exposed now.

Would be good to make `warning` matches genuinely advisory: keep them visible in the run record, but don't let them alone push a node to `semantic_fail`. Save that for `critical` severity, validator failures, and critical anomalies, same as today.

This would also make it reasonable to keep ambiguous signatures like the back-reference ones around without them causing false failures.

AaryanPuri:
Closes #46.

Right now, any match from the signature registry fails a node, no matter what severity it's marked as. A warning hit (like an ambiguous placeholder match) ends up failing the node exactly the same way a critical one does, which isn't really what "warning" is supposed to mean. This came up during review on PR #43.

Fixed the core logic in session.py so only critical severity signals actually flip a node's status now. Warnings still show up on the run for visibility, they just don't fail anything on their own anymore.

That change turned out to have a real side effect though: the learned-signature feature (where a pattern gets approved and then auto-flags future runs without needing an LLM call) was quietly depending on the old behavior, since candidates default to warning severity. So I updated approve_candidate and approve_candidate_shared in candidate_store.py to bump severity to critical the moment a human approves a candidate. Makes sense either way, once you've confirmed a pattern is real, it's not a hedge anymore. Kept the original suggested severity in the metadata for reference.

Updated the DB-flow tests to match, and added a couple of small focused tests just for the severity gate itself so this doesn't regress quietly again.

### DIFF
--- a/src/argus/candidate_store.py
+++ b/src/argus/candidate_store.py
@@ -163,7 +163,11 @@ def approve_candidate(candidate_id: str) -> dict[str, Any] | None:
         "category": cand["proposed_category"],
         "pattern": cand["pattern"],
         "match_strategy": cand["match_strategy"],
-        "severity": cand["severity"],
+        # Approval promotes severity to critical: a candidate is a hedge
+        # (LLM-suggested, unconfirmed), but a developer explicitly approving
+        # it is a confirmed, recurring pattern — no longer ambiguous, so it
+        # should flag future runs on its own rather than sit as advisory-only.
+        "severity": "critical",
         "description": cand["description"],
         "source": "learned",
         "metadata": {
@@ -174,6 +178,7 @@ def approve_candidate(candidate_id: str) -> dict[str, Any] | None:
             "framework_specific": None,
             "original_pattern": cand.get("original_pattern"),
             "generalized": cand.get("generalized", False),
+            "suggested_severity": cand["severity"],
         },
     }
     custom["signatures"].append(new_sig)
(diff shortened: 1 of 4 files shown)


VaradDurge (approved):
Root-cause fix for the severity-gating inconsistency (mirrors has_tool_failure). Correctly handles the learned-candidate side-effect by promoting approved candidates to critical. Verified: SP-022 back-reference now passes; critical refusals still fail; 132 tests pass. Merging; detection re-triage tracked separately.

PR #58, opened 2026-08-30
test: use sys.executable in TestCLI subprocess calls

abhisheksharma001:
## What & why

`tests/test_e2e_langgraph.py::TestCLI` ran `python3 -m argus.cli.main` via `subprocess`. On any machine where the venv interpreter isn't the `python3` on PATH, `argus` isn't importable in that subprocess and both tests fail — reproduced on `master` while reviewing #48. `sys.executable` pins the subprocess to the interpreter running pytest.

Tracks `visual/PRD.md` US-6.1.

## Type of change

- [x] Bug fix

## Checklist

- [x] I have read the Contributing guide.
- [x] CLA signed.
- [x] Open-source core only — no `cloud/` or `supabase/` changes.
- [x] `pytest tests/test_e2e_langgraph.py::TestCLI` → 2 passed.
- [x] `ruff check .` passes.
- [x] No public API or behavior change; no docs needed.
- [x] Label: `bugfix`.

## Notes for reviewers

`ruff format --check` flags this file, but that is pre-existing on `dev` (verified by stashing); left untouched to keep the diff to the fix.

PR #59, opened 2026-08-30
docs: specify the status vocabulary and node → run roll-up

abhisheksharma001:
## What & why

The 50-agent stability review (`ARGUS_STABILITY_REVIEW.md`, P0 #2) found the status taxonomy undocumented: consumers observed `semantic_fail` on steps but `silent_failure` on the run and had no spec to write assertions against. This PR writes that spec and makes the type system enforce it.

- **`docs/STATUS.md`** — every `NodeEvent.status` and `RunRecord.overall_status` value, who assigns it, whether it counts against the run, the roll-up order (`crashed` > `interrupted` > `silent_failure` > `clean`), the `retried`/`skipped` exclusion rule, what `argus check` treats as failing, and the neighbouring vocabularies (`Severity`, correlator `event_type`) that look like statuses but aren't.
- **`models.StepStatus` / `models.RunStatus`** `Literal` aliases, applied to `NodeEvent.status`, `RunRecord.overall_status`, and the three `session.py` sites that produce them. mypy now rejects an undocumented status. Error count unchanged from `dev` (233 → 233).
- **Stale lists corrected:** `CLAUDE.md` step 7 (was missing `degraded_input`, `retried`, `skipped`); `models.py` comment (was missing `interrupted`). README "Detection Layers" links to the doc.

Documented, **not** changed: run-level `semantic_fail` appears in `check.UNCLEAN_OVERALL_STATUSES` and `website/lib/types.ts` `RunStatus` but is never produced — node `semantic_fail` rolls up to `silent_failure`. Left as tolerant superset; flagged in the doc.

Tracks `visual/PRD.md` US-1.1. Next in sequence: US-1.2 normalized `findings[]`, which depends on this vocabulary.

## Type of change

- [x] Docs (+ type annotations, no runtime change)

## Checklist

- [x] Contributing guide read; CLA signed.
- [x] Open-source core only.
- [x] `pytest` — 795 passed; the 2 `TestCLI` failures are the pre-existing env issue fixed in #58 (this branch is off `dev`).
- [x] `ruff check .` passes. `mypy src/argus` error count unchanged.
- [x] Docs updated: `CLAUDE.md`, `README.md`, new `docs/STATUS.md`.
- [x] Label: `documentation`.

## Notes for reviewers

Two things worth a decision, neither blocking:
1. Should `semantic_fail` be dropped from `UNCLEAN_OVERALL_STATUSES` / TS `RunStatus` since it's dead? Trivial follow-up if yes.
2. `retried` only applies when the final loop iteration is `pass`; if it fails, earlier iterations keep their own status and each counts. That's current behavior and now documented — confirm it's intended.


VaradDurge:
Reviewed after merging current `master` into this branch — same stacked-on-old-`dev` situation as #58. GitHub’s 20-file / ~3k-line diff was leftover time-travel + `visual/` that already landed in #58. Unique remaining change is 5 files / +115 −10:

- `docs/STATUS.md` (new)
- `StepStatus` / `RunStatus` Literals on `NodeEvent.status`, `RunRecord.overall_status`, and the assignment sites in `session.py`
- stale list fixes in `CLAUDE.md` and the README Detection Layers link

Checked against `_finalize` roll-up (`crashed` > `interrupted` > `silent_failure` > `clean`), `_apply_loop_retries` (`retried` only when the final loop iteration is `pass`), skipped unchosen branches excluded from roll-up, and `argus.check.evaluate_run`. No runtime behavior change.

Non-blocking follow-ups (do not hold this):

1. Table says a blocking validator sets `fail`; in `session.py` a blocking validator on a previously-`pass` node actually sets `semantic_fail`. Both still roll up to run `silent_failure` and fail `argus check`.
2. `_finalize` does not explicitly test `status == "fail"`; `fail` is assigned only when inspection already has silent/tool failure, so the documented roll-up still holds. `argus check` also treats active `fail` as unclean even if overall were somehow `clean`.
3. Assigned-by names: retries are `_apply_loop_retries`, skipped nodes are inlined in `_finalize` (not `_mark_retried_iterations` / `_record_skipped`).
4. Dead run-level `semantic_fail` in `check.UNCLEAN_OVERALL_STATUSES` and `website/lib/types.ts`: keep as a tolerant superset for now, as the doc already says. Drop in a follow-up if we want the TS type to match `models.RunStatus`.

Approving and squash-merging once CI is green.

VaradDurge (approved):
LGTM. Unique remaining change after merging master is STATUS.md + Literal types. No runtime behavior change. Non-blocking nits on the PR comment.

PR #62, opened 2026-08-30
feat: add argus ignore — project-level signature suppressions

abhisheksharma001:
## What & why

`ARGUS_STABILITY_REVIEW.md` P1 #4: signatures are over-eager on legitimate content (`NL-002` on the string `"none"`, `RF-001` on markdown bullets, `BA-005` on flat scalars) and "the current incentive is to contort *content* to satisfy the detector — backwards. Need per-node or per-project signature config (`argus ignore NL-002 --node draft_hook`) persisted in `.argus/config`." This is that command.

```
argus ignore NL-002                    # everywhere in this project
argus ignore RF-001 --node draft_hook  # only on one node
argus ignore --list
argus ignore NL-002 --remove
```

- Stored in `<project>/.argus/config.json` → `"suppressions": [{"id","node"}]`. Commit it to share.
- **Not dropped.** `session.py` partitions `semantic_signals` and `anomaly_signals` *before* status is decided. Suppressed hits move to new `NodeEvent.suppressed_signals` / `suppressed_anomalies` (persisted + deserialized) so `argus stats` and findings can still count them; the node status, run verdict, and CI gate no longer see them.
- Works for heuristic ids (`PH-*`, `NL-*`, `RF-*`, …) and anomaly ids (`BA-*`).
- `argus doctor` shows a `suppressions` row. README + CLAUDE.md updated.

Independent of the #59/#60/#61 stack (branched from `dev`). Once #60 lands, one follow-up line emits suppressed hits as `Finding(suppressed=True)`.

Tracks `visual/PRD.md` US-1.3.

## Type of change

- [x] New feature (new CLI command, two additive `NodeEvent` fields)

## Checklist

- [x] Contributing guide read; CLA signed.
- [x] Open-source core only.
- [x] Tests: 9 new in `tests/test_suppressions.py`, incl. a real-session integration test (`{"summary": "TBD"}` → PH-007 → `semantic_fail`; after `argus ignore PH-007` → `pass`, hit recorded in `suppressed_signals`, round-trips through storage). `pytest` → 804 passed; the 2 `TestCLI` failures are fixed in #58.
- [x] `ruff check .` passes. `mypy src/argus`: 233 → 236 — the three new lines are the same class every CLI module already carries (`rich.*` import-not-found in a stub-less env ×2, `Untyped decorator` on the typer command ×1); no new type errors in `suppressions.py`, `session.py`, or `storage.py`.
[...]

PR #63, opened 2026-08-30
fix(ui): make /runs/<id> deep links work; remove dead header buttons

abhisheksharma001:
## What & why

Found while driving the dashboard end-to-end in a real browser.

**1. Deep links to runs were broken.** `website/app/runs/[id]/page.tsx` uses `generateStaticParams() → [{id:'_'}]`; the static export bakes that placeholder into the page, so `RunDetailRedirectClient` received `id="_"` at runtime for *every* URL and redirected to the bare run list. Pasting a run link into Discord/a PR never opened the run. Fix: read the id from `window.location.pathname` (fallback to the prop), then forward to the existing `/?run=<id>` view. Verified live: `http://localhost:7842/runs/<id>` now renders the full detail page.

**2. Dead header buttons.** `Filters` and `Last 1h` in the Runs header had no `onClick` — two controls implying time filtering that doesn't exist (the All/Clean/Failed/Semantic/Interrupted tab row is the real filter). Consequence in practice: header says "871 runs · 44 failed", user sees a page of green and a "Last 1h" label, and assumes the failures are filtered out. Removed both; the **clean/failed counts in the header are now buttons** that activate the matching filter tab, so the shortest path from "44 failed" to the failed runs is one click. Verified live: clicking the count flips the active tab.

`src/argus/ui_dist/` rebuilt with `scripts/build_ui.sh` and committed, per the script's convention ("Commit it with your release") — that's most of the diff; the source change is 2 files.

## Type of change

- [x] Bug fix

## Checklist

- [x] Contributing guide read; CLA signed.
- [x] Open-source core only.
- [x] `npm run build` clean; behavior verified in a live browser (deep link renders detail; count click activates filter).
- [x] Python side untouched; `pytest`/`ruff` unaffected.
- [x] Label: `bugfix`.

## Notes for reviewers

- A time-range filter would be genuinely useful — happy to build a real one (`1h/24h/7d/all` acting on `started_at`) as a follow-up; this PR only stops the UI from claiming one exists.
- Remaining dashboard observations (0-step run clutter, tabs missing ARIA roles, Approvals count not surfaced on Runs) are logged in `visual/PRD.md` scope on `dev`; can file as issues on request.


VaradDurge:
Hey @abhisheksharma001 — rebased this onto current `master` (it was `BEHIND`, 11 commits back). Pushed to `fix/dashboard-deeplink` (force-update). **Not merged — leaving that to you.** The UI fix itself I reviewed and it's correct as-is; no code changes from me.

### Rebase
Same divergence as #62: 10 of the 11 commits (the replay/state-patch stack + `visual/` docs) were already on master identically, so I rebased with `--onto` to keep only the real fix commit (`4c5050f`). Applied cleanly, **no conflicts**. Now **1 commit / `MERGEABLE`** (was 78 files / +3459 of divergence noise; the real change is the 2 `.tsx` files + the rebuilt bundle).

### What I checked
- **Deep-link fix** (`RunDetailClient.tsx`) — reading the id from `window.location.pathname` with the prop as fallback + `encodeURIComponent` is sound; the regex handles the trailing slash.
- **Header counts** (`RunListPanel.tsx`) — verified the "N failed" button isn't a count/filter mismatch: `counts.failed` and the `crashed` filter use the **same** predicate (`overall_status === 'crashed' || 'silent_failure'`), and the "Failed" tab is `id: 'crashed'`, so the number and the filtered list agree. Good.
- **Committed `ui_dist/` bundle** — master's only `website/` change since your base is `lib/types.ts` (+17), and it's **pure type declarations** (`Finding`/`FindingSource`, erased at build → zero runtime JS). So your committed bundle is runtime-identical to a fresh build on master; I did **not** need to rebuild it.

### Minor note (non-blocking)
`website/package-lock.json` drops 36 lines, all `"libc": ["glibc"|"musl"]` platform-hint metadata on optional native deps — lockfile churn from a different npm version, no package/version/integrity removed. Harmless; mentioning only so it's not a surprise in the diff.

`MERGEABLE`, gated on your review.

VaradDurge (approved):
Reviewed: deep-link fix reads the real id from the path with a safe fallback; dead header controls removed; clickable clean/failed counts share the same predicate as their filter tabs (no count/filter mismatch). Rebased onto master, bundle verified. LGTM.

PR #64, opened 2026-08-31
fix(check): grade and report the exact run

### LINKED ISSUE #55: URGENT: argus check must grade a specific run and print the file
Parent: #52

## Problem

`argus check last` grades the **newest file** in `.argus/runs/` (`last_run_id()`).

That is a guess. It breaks when:

- CI `cd`s into a subfolder → “No runs found”
- an old failed run is still on disk → you grade the ghost
- two jobs share a machine → newest ≠ yours

## What to do

1. Accept a run from the environment: `ARGUS_RUN_ID` (and/or the id the invoke just wrote).
2. Print the exact path graded, e.g. `checked /path/to/.argus/runs/<id>.json`.
3. Keep `check last` but treat it as “newest file — easy to get wrong.” Document that.

Do not rewrite `evaluate_run()`. Only change how the CLI picks **which** record to load.

## Done when

CI can `argus check $ARGUS_RUN_ID` and the log shows which file was read. Tests cover: env id wins over `last`, missing id still errors clearly, printed path matches the file loaded.

## Files

- `src/argus/cli/cmd_check.py`
- `src/argus/storage.py` (`last_run_id`)
- `tests/test_cmd_check.py`

xianjianlf2:
## Summary

- select runs in deterministic order: explicit CLI id, `ARGUS_RUN_ID`, then newest-file fallback
- expose the exact resolved run JSON path and print it in `argus check` output
- document why CI should prefer exact ids over `check last`
- preserve the existing `evaluate_run()` behavior unchanged

## Verification

- `uv run --frozen pytest tests/` — 759 passed, 5 skipped, 2 xfailed
- `uv run --frozen ruff check src/` — passed
- focused check/storage suite — 36 passed

`mypy src/argus` currently reports 181 pre-existing errors on `master`; the full test and Ruff gates pass.

Closes #55


abhisheksharma001:
**Review (medium effort)**

1. `check_run()` resolves the run path twice — once directly via `resolve_run_path(target)`, again inside `load_run(target)`. Each call does an unfiltered `rglob(".argus/runs")` walk, so this doubles filesystem I/O on every `argus check` invocation. It also opens a small TOCTOU gap: if a run file rotates or a new file lands between the two resolutions, the path printed as `checked` can diverge from the record actually graded.

   Suggest: resolve once, pass the resolved path into `load_run`/`load_run_text` instead of re-resolving.

2. `argus check last` is silently overridden by a stale `ARGUS_RUN_ID` env var. `target = run_id if run_id not in (None, "last", "run") else None` treats the literal `last` arg the same as no arg, so if `ARGUS_RUN_ID` is set (e.g. left over from an earlier CI step in the same job), `argus check last` grades the env-selected run instead of the newest one — even though the user explicitly typed `last`. Confirmed intentional via `test_check_argus_run_id_wins_over_last`, but worth a second look: a CI script that always runs `argus check last` won't realize a prior step's exported `ARGUS_RUN_ID` is still shadowing it.

Everything else checks out — precedence tests cover the intended order, docs match implementation, exception handling preserved.

VaradDurge (approved):
Reviewed against issue #55: deterministic run selection (explicit id > ARGUS_RUN_ID > newest-file fallback), prints the exact graded path, evaluate_run() untouched. resolve_run_path refactor is behavior-preserving; invalid id exits 1 with a clear message. Verified live (explicit id, env id, bogus id) and ran the check/storage suites (40 passed) plus ruff clean on a local merge with master. LGTM.

PR #61, opened 2026-08-30
feat(check): add --format json and --fail-on to argus check

abhisheksharma001:
> **Stacked on #60** (findings) ← #59 (vocab). Base is `feat/run-findings`; retarget to `master` once the stack below merges.

## What & why

`ARGUS_STABILITY_REVIEW.md` P0 #1: "No machine-readable verdict. A CI gate needs `argus check <run-id> --format json` (or `--fail-on silent_failure`) with a meaningful exit code. Today every harness must hand-parse undocumented JSON internals." This closes it.

```
$ argus check last --format json
{
  "run_id": "20260830-…",
  "schema_version": "2",
  "overall_status": "silent_failure",
  "passed": false,
  "fail_on": null,
  "first_failure_step": "search",
  "root_cause_chain": ["search"],
  "failing_nodes": ["search"],
  "reasons": ["search: fail", "search: silent_failure", "search: missing_fields (documents)"],
  "findings": [ { "id": "…", "node": "search", "type": "missing_field", "severity": "critical",
                  "reason": "Field `documents` required downstream was not set by `search`.",
                  "source": "heuristic", "field_path": "documents", … } ]
}
$ echo $?
1
```

- `--format json`: one object on stdout, nothing else. Exit codes unchanged (0 pass / 1 fail). Usage errors exit 2 and emit `{"error": …}` under json so stdout is always parseable.
- `--fail-on crashed,interrupted,silent_failure`: gate only on those run statuses; default = any non-clean. Validated against `docs/STATUS.md`'s roll-up set (`semantic_fail` is intentionally not accepted at run level — it never occurs; see #59).
- `check_payload()` exported for the pytest plugin / exporters.
- Text mode unchanged except a dim `fail-on …` line when the flag is set.

Tracks `visual/PRD.md` US-2.1. Unlocks the GitHub Actions example (US-5.3) and the webhook exporter body (US-5.1).

## Type of change

- [x] New feature (additive CLI options)

## Checklist

- [x] Contributing guide read; CLA signed.
- [x] Open-source core only.
- [x] Tests: 4 new in `tests/test_cmd_check.py` — json fail shape + exit 1, json clean exit 0, `--fail-on` filters both ways, unknown value → exit 2 with json error. `pytest` → 803 passed; the 2 `TestCLI` failures are fixed in #58.
- [x] `ruff check .` passes; `mypy src/argus` error count unchanged from base.
- [x] Docs: README CLI block; `argus check --help` shows both options.
- [x] Label: `enhancement`.

[...]

PR #65, opened 2026-09-01
feat: add chat_response and code_generation behavior profiles

### LINKED ISSUE #24: Add behavior profiles for chat and code-generation nodes
## What

The anomaly detector (`src/argus/anomaly_detector.py`) uses behavior profiles to set expected output characteristics per node type. Current profiles:

- `structured_json`
- `retrieval_result`
- `reasoning_chain`
- `tool_output`
- `free_text`

Two common pipeline node types are missing:

1. **`chat_response`** — Conversational LLM nodes that produce user-facing replies. Expected: moderate length (50–5000 chars), low repetition, no placeholder phrases, higher info density than free_text.
2. **`code_generation`** — Nodes that emit code snippets or scripts. Expected: may contain repeated structural patterns (indentation, braces) that shouldn't trigger repetition anomalies, higher tolerance for special characters, length range (10–100,000).

Without these profiles, ARGUS auto-infers behavior which can trigger false positives (e.g., code indentation flagged as repetition).

## How

1. Add the new profile dicts to `BEHAVIOR_PROFILES` in `anomaly_detector.py`
2. Follow the existing profile structure (keys: `min_keys`, `expects_nested`, `min_info_density`, `length_range`, etc.)
3. Add tests that run the anomaly detector on representative chat/code outputs and verify no false anomalies fire

## Scope

~20 lines of profile config + tests. Good way to learn how the detection pipeline works.

abhisheksharma001:
Closes #24.

## What
Adds two missing entries to `BEHAVIOR_PROFILES` in `anomaly_detector.py`:

- `chat_response` — 50-5,000 chars, `min_info_density` 0.30 (higher than a generic text profile), no nesting expected.
- `code_generation` — 10-100,000 chars, `min_info_density` 0.10 (code repeats keywords like `def`/`self`/`return` more than prose, so needs a lower floor than `chat_response`).

Without these, nodes of these types fell through to auto-inference and could misfire - e.g. code indentation/structure flagged as low info density.

## Verification
Checked whether the repetition check (BA-002) would also need touching for code's repeated braces/indentation - it doesn't: `_check_repetitive_filler` tokenizes with an alnum-only word regex, so punctuation and whitespace are already stripped before the repetition ratio is computed. No core check logic changed, just the two profile dicts.

Added 3 tests: a clean chat reply, a clean function, and a multi-method class (to specifically exercise the repeated-braces/indentation case) - all confirm zero anomalies fire under the new profiles.

`pytest tests/test_anomaly_detector.py` - 36 passed
`ruff check` - clean


VaradDurge:
Reviewed after merging current `master` into this branch — same stacked-on-old-`dev` situation as #58–#61. GitHub’s 20-file / ~3k-line diff is leftover time-travel + `visual/` that already landed in #58. Unique remaining change is 2 files / +60:

- `BEHAVIOR_PROFILES["chat_response"]` — 50–5,000 chars, `min_info_density` 0.30, no nesting
- `BEHAVIOR_PROFILES["code_generation"]` — 10–100,000 chars, `min_info_density` 0.10
- 3 tests: clean chat reply, clean function, indented multi-method class (BA-002 does not fire)

Matches #24’s How/Scope (profile dicts + no-false-positive tests). BA-002 already tokenizes `[a-zA-Z0-9]+`, so braces/indent were never the repetition problem. The real win is length + density when the profile is selected via `BehaviorConfig`. Local `tests/test_anomaly_detector.py`: 36 passed.

Non-blocking follow-ups (do not hold this):

1. `infer_behavior_type` still never returns `chat_response` or `code_generation`. Without `behavior_type=` / `node_behaviors=`, a chat or code node still lands on `classification` / `detailed_text` / `structured_json`. The profiles only apply when configured. A later PR can add inference (e.g. fenced code / conversational length band) if we want #24’s “auto-infer misfire” story closed for unconfigured graphs.
2. Tests are happy-path only. A one-liner that a 10-char chat reply still trips BA-001 under `chat_response` would pin that the tighter floor is live.
3. `chat_response` 5,000-char cap is the advertised range; BA-001 explosion is `hi * 3` (15k), so long RAG answers will not warn until well past 5k. Fine, same slack as the other profiles.

Approving and squash-merging once CI is green.

VaradDurge (approved):
LGTM. Unique remaining change after merging master is two BEHAVIOR_PROFILES entries plus three no-false-positive tests. Closes #24 as specified. Non-blocking nits on the PR comment.

PR #69, opened 2026-09-03
fix(ui): label the failure types and behavior profiles the dashboard was dropping

VaradDurge:
First of three stacked PRs redesigning the dashboard. This one is small and
independently correct — it fixes real data bugs and can merge on its own.

## What was wrong

The dashboard mirrors Python enums by hand in TypeScript, and the two had drifted:

- `empty_output` (#36) and `json_in_string` (Rule 17, #34) had no entry in
  `FAILURE_META`, so `getFailureMeta()` fell through to the grey **"Unknown"**
  chip. Two headline detection rules rendered as nothing.
- `chat_response` and `code_generation` (#65) were missing from `BEHAVIOR_LABELS`,
  so they rendered as raw snake_case.
- `state_patch` (#29), `coverage_summary` and `schema_version` exist on the Python
  `RunRecord` but not the TypeScript one, so the UI could not read them.

## The fix that matters

Rather than patch the four values and wait for the next drift, `tests/test_ui_parity.py`
asserts the invariant: every `failure_type`, behavior profile and `StepStatus` the
Python side emits must have a UI label. It fails with the offending names listed.

It failed RED on exactly those four before the fix.

## Test plan

- [x] `pytest tests/test_ui_parity.py` — 3 passed
- [x] `tsc --noEmit` — 0 errors

### DIFF
--- a/docs/superpowers/plans/2026-09-03-dashboard-redesign.md
+++ b/docs/superpowers/plans/2026-09-03-dashboard-redesign.md
@@ -1,0 +1,1417 @@
+# Dashboard Refactor + Redesign Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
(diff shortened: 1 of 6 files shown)


VaradDurge:
Folded into #71. The three stacked UI PRs are now a single PR against `master` (`ui-spec-redesign`). This branch’s commits are already in that PR.

PR #70, opened 2026-09-03
refactor(ui): dual-theme token system and ~500-literal cleanup

VaradDurge:
Second of three stacked PRs. **Base is `ui-data-parity` (#69)** — review that first.

## Why this was needed

The dashboard was **dark-only**: `:root { color-scheme: dark }`, no
`prefers-color-scheme`, no `[data-theme]`, no toggle. And ~500 hardcoded hex
literals lived inside `.tsx` files (worst offenders: `StepInspector.tsx` with 69,
`approvals/page.tsx` with 37).

A hardcoded `#ef4444` cannot respond to a theme change, so de-hardcoding was a
prerequisite for light mode, not a nicety.

## What changed

- Full dual-theme token set — light, `prefers-color-scheme: dark`, and
  `[data-theme]`, so an explicit user choice beats the OS in **both** directions.
- A **bridge block** aliases the old token names (`--card` → `--panel`, etc.) so
  hundreds of existing Tailwind classes keep working instead of a big-bang rewrite.
- Theme toggle with a blocking inline script, so a stored theme applies before
  first paint — no flash.
- **~500 literals → 0.** `npm run check:literals` ratchets it: files graduate onto
  a guarded list as they are converted. Only exception is one modal scrim, which is
  deliberately black in both themes and documented as such.
- Two oversized files split while being converted: `StepInspector` 795 → 526 lines,
  `approvals/page` 1027 → 274.

Dropped the Geist/JetBrains webfonts for the spec's system stacks — zero webfont
requests now.

## Test plan

- [x] `npm run check:literals` — 0 literals across 73 files
- [x] `tsc --noEmit` — 0 errors
- [x] `npm run lint` — clean
- [x] `npm run build` — 14/14 routes
- [ ] **Visual check in light mode** — the one thing not yet verified by a human


VaradDurge:
Folded into #71. The three stacked UI PRs are now a single PR against `master` (`ui-spec-redesign`). This branch’s commits are already in that PR.

PR #71, opened 2026-09-03
feat(ui): Argus Instrument dashboard redesign

VaradDurge:
## What & why

One PR for the dashboard redesign (was stacked as #69 + #70 + #71). All commits are on `ui-spec-redesign`; nobody else has pushed to these branches.

Closes #69, supersedes #70.

## What changed

**1. Data parity (was #69)**
- Label `empty_output` and `json_in_string` (they rendered as **Unknown**)
- Label `chat_response` / `code_generation` behavior profiles
- Mirror `state_patch`, `coverage_summary`, `schema_version` on the TS `RunRecord`
- `tests/test_ui_parity.py` fails if Python emits a type the UI cannot name

**2. Dual-theme tokens (was #70)**
- Light / `prefers-color-scheme: dark` / `[data-theme]` with a no-flash toggle
- Bridge aliases so existing Tailwind classes keep working
- ~500 hex literals → 0 (`npm run check:literals`)
- Split `StepInspector` and the approvals page while converting

**3. Spec shell, local-only OSS, execution graph (was #71)**
- Dashboard is local `argus ui` only — removed hosted Supabase / Google login from `website/`
- Spec component layer (chips, tiles, table, banners, graph nodes)
- Interactive execution graph (drag / pan / zoom / inspect, ROOT CAUSE ribbon)
- CLI cloud (`argus login`, `cloud/`) is untouched

Home first-load JS 194 kB → 132 kB.

## Type of change

- [x] New feature
- [x] Refactor / internal
- [x] Bug fix

## Checklist

- [x] Open-source core only — `cloud/` / `supabase/` not modified
- [x] `pytest tests/test_ui_parity.py` — 3 passed
- [x] `tsc --noEmit`, `npm run lint`, `npm run build` clean
- [x] `npm run check:literals` — 0 literals
- [ ] Visual review in both themes
- [ ] Graph interaction — drag / pan / zoom / inspect

## Notes for reviewers

`src/argus/ui_dist/` is a tracked build artifact (`scripts/build_ui.sh`). `argus ui` serves that, not `website/out`.


abhisheksharma001 (changes requested):
Thanks @VaradDurge — this is a big, well-organized redesign and most of it checks out: auth/Supabase removal is clean, the token architecture with the alias bridge and no-flash inline script is solid, `pytest` (864 passed), `tsc --noEmit`, `next lint`, and `next build` all pass locally, and `cloud/`+`supabase/` are untouched. Four things need fixing before merge:

1. **`src/argus/ui_dist/` is stale.** The last commit (`7bbaec7`, the spec IDE shell) never rebuilt the artifact — the committed dist has no `Failure hotspots`, `api/hotspots`, or `/fix` strings, while a fresh build from this branch does. Since `argus ui` serves this directory, users would get the old dashboard. Please run `bash scripts/build_ui.sh` and commit. Longer-term, a CI job that rebuilds and fails on diff would prevent recurrence (committing the artifact already bloats every diff; staleness defeats its purpose).

2. **`npm run check:literals` crashes**: `GUARDED` in `scripts/check-no-literals.mjs` still lists deleted files (`components/Sidebar.tsx`, `Topbar.tsx`, `app/report/*`) → `ENOENT` at line 84, so the guard currently checks nothing. Prune the list and add an `existsSync` filter.

3. **`/api/hotspots` doesn't exist.** `HotspotMatrix.tsx:55` fetches it, gets a 404, and falls back to fetching up to 40 full run records client-side; meanwhile `src/argus/hotspots.py` + its test are dead code. Either wire the route in `cmd_open_ui.py` or drop the Python module. Note the two paths also diverge (500-run all-status cap vs 40 failed-run cap).

4. **`ExecutionGraph.mapStatus` doesn't handle `interrupted`** (a real `StepStatus`) — it falls through to `default: 'pass'`, so interrupted nodes render green. Add a case, and consider extending the parity test to cover this map, not just `StatusBadge`.

Nits: 34 hex literals remain in unguarded `lib/run-utils.ts` (leaks dark-palette grays into light theme via `RunHeader`/`StepRow`); `aria-selected` on a button in `RunFilterBar.tsx:83`; `useMemo` dep warning in `PipelineTab.tsx:175`; the parity test's behavior-profile list is hardcoded so new Python profiles won't trip it; dead `googleusercontent.com` entry in `next.config.mjs`; and home first-load JS measures 138 kB here vs the claimed 132 kB. Also worth noting CI runs only the CLA check on PRs — items 1–3 would all have been caught by a pytest+build+check:literals workflow.

VaradDurge (commented):
Would request changes (GitHub blocks REQUEST_CHANGES on the author's own PR). Independently reviewed the source diff (website/, `cmd_open_ui.py`, `hotspots.py`, tests). Detection/inspector/check are untouched — this is UI/CLI serving only. Not merge-ready.

## Blockers

1. **`src/argus/ui_dist/` is stale.** Last commit (`7bbaec7`, spec IDE shell) never rebuilt the artifact. The committed homepage chunk has no `Failure hotspots` / `api/hotspots` / `/fix` strings. `argus ui` serves this directory, so users get the previous dashboard. Rebuild with `bash scripts/build_ui.sh` and commit. A CI job that rebuilds and fails on diff would stop this recurring.

2. **`npm run check:literals` crashes.** `GUARDED` in `website/scripts/check-no-literals.mjs` still lists deleted files (`components/Sidebar.tsx`, `Topbar.tsx`, `app/report/*`). `readFileSync` throws `ENOENT`, so the guard checks nothing. Prune the list and skip missing paths.

3. **`/api/hotspots` is not wired.** `HotspotMatrix.tsx` fetches it, 404s, then fans out up to 40 full run GETs. `src/argus/hotspots.py` is unused. Either add the route in `cmd_open_ui.py` or delete the Python module. The two paths also disagree (Python: 500-run all-status cap + tag filter; client: 40 failed runs).

4. **`ExecutionGraph.mapStatus` maps `interrupted` to green `pass`.** `interrupted` is a real `StepStatus` and `StatusBadge` already handles it. `default: return 'pass'` is wrong. Add a case. Extend `test_ui_parity.py` to cover this map, not only `StatusBadge`.

## Also fix before merge

5. **Copy-fix leaks recorded state.** `website/lib/fix-prompt.ts` calls `/api/runs/:id/fix` with no `sanitized` query. `build_fix_prompt_for_record(..., sanitized=False)` is the default, so the dashboard Copy action embeds input/output values. Either pass `sanitized=1` by default and offer a "include values" toggle, or make the default explicit in the UI.

6. **Theme tokens are incomplete.** `website/lib/run-utils.ts` still has hex constants (`#10b981`, `#6b7280`, `#374151`, …). `RunHeader` and `StepRow` import from it, so light theme still gets dark-palette grays. It is also unguarded by `check:literals`.

7. **`uv.lock` (≈3k lines) is new vs `master` and unrelated to this UI change.** Drop it unless the repo is switching to uv in this PR.

8. **Local-only claim is incomplete.** Auth/Supabase client code is gone, but `website/app/layout.tsx` still ships hosted marketing metadata (`canonical` / `og:url` = `https://arguslabs.in`, `robots: index`). `next.config.mjs` still allows `*.googleusercontent.com`. If `website/` is now only `argus ui`, strip the public-site SEO and the leftover image host.

## Nits (non-blocking)

- `empty_output` is labeled category `Tool` in `failure-labels.ts`. It is a state-update no-op, not a tool I/O failure.
- `ToolFailure` in `types.ts` omits `timeout_adjacent` / `suspiciously_fast` / `latency_quality_mismatch` even though `FAILURE_META` has them.
- `test_every_behavior_type_has_a_ui_label` hardcodes the Python profile list; a new profile will not fail the test.
- `RunFilterBar.tsx` uses `aria-selected` on a `<button>` that is not a tab.
- `PipelineTab.tsx` `useMemo` at the `segmentEvents` call is missing a dependency (lint warning).
- `docs/superpowers/plans/2026-09-03-dashboard-redesign.md` is 1417 lines of process notes. Prefer keeping that out of the product PR.
- CI on this PR is CLA-only. Items 1–3 would have been caught by pytest + `next build` + `check:literals`.

The token work, auth removal, and `empty_output` / `json_in_string` labels are the right direction. Rebuild dist, wire or drop hotspots, fix `mapStatus`, and do not merge the leftover lockfile/SEO.

abhisheksharma001:
## Review

- 🔴 bug `website/components/run-detail/ExecutionGraph.tsx:160` — sentinel filter lost. `useMemo(() => run.graph_node_names ?? [], [run])` keeps `__start__`/`__end__` (`watcher.py:227` does raw `list(graph.nodes.keys())`); the old code filtered them. Every LangGraph run now renders phantom sentinel nodes as "not reached". Fix: re-add `.filter((n) => !n.startsWith('__') && n !== 'START' && n !== 'END')`.
- 🔴 bug `website/scripts/check-no-literals.mjs:14` — GUARDED still lists deleted `components/Sidebar.tsx` (and `components/Topbar.tsx`); `readFileSync` throws ENOENT so `npm run check:literals` crashes at branch tip, contradicting the PR checklist. Fix: drop dead entries or skip missing files.
- 🟡 risk `src/argus/hotspots.py` — dead code. `/api/hotspots` is never registered in `cmd_open_ui.py`, so `HotspotMatrix.tsx` always uses its ≤40-run client-side fallback; `finding_index()` has no production caller. Likewise `RunSummary.origins`/`finding_nodes` (`website/lib/types.ts:354`) are never emitted by `_list_runs` (`cmd_open_ui.py:717-740`), so origin/node filter chips collapse to `first_failure_step` values. Fix: wire `finding_index` into summaries + add the `/api/hotspots` route, or delete the Python module and TS fields.
- 🟡 risk `website/lib/workspace.tsx:92` — `fetchRuns` keeps `prev` when only `run_id`+`overall_status` match, so live-tail (6s poll) never shows in-flight progress (step_count/duration_ms on an `interrupted` run) until status flips. Fix: include `step_count`/`duration_ms` in the equality check or always `setRuns(data)`.
- 🟡 risk `src/argus/hotspots.py:40` — `_matches_tag`: `tag in str(tags)` is substring on the dict repr, and `value in tags.get(key, [])` is substring when tag value is a string (`env:pro` matches `{env: "prod"}`). Fix: exact membership / `isinstance(str)` comparisons.
- 🟡 risk `website/components/RunDetailPanel.tsx:55` — `useFixPrompt(runId ?? '', undefined, { autoload: true })` fires `GET /api/runs/<id>/fix` (→400) for every opened run including clean ones; render is gated by `canFix` but the fetch is not. Fix: `autoload: hasFindings`.
- 🔵 nit `src/argus/cli/cmd_open_ui.py:918` — `(qs.get("node", [None])[0] or None) or None` redundant second `or None`.
- 🔵 nit `website/components/ThemeToggle.tsx` — unused; `IconRail.tsx` implements its own separate `ThemeButton` (two unsynced theme UIs). Fix: delete `ThemeToggle` or render it and drop `ThemeButton`.
- 🔵 nit `website/components/run-detail/ExecutionGraph.tsx:96` — `mapStatus` `default: return 'pass'` maps any future/unknown `StepStatus` to green. Fix: return `'skipped'` for unknown.
- 🔵 nit `website/components/HotspotMatrix.tsx:44` — effect deps reduced to `[runKey]` with eslint-disable: run whose status flips without id list changing never refreshes matrix. Fix: depend on `runs` (or status signature).
- 🔵 nit `tests/test_ui_parity.py:26` — `failure_type="([a-z_]+)"` only catches literal kwargs; `failure_type=CONST` or f-strings evade the guard. Fix: also scan a shared constants module if one is added.
- ❓ q: `uv.lock` (+3028, new file) with no `pyproject.toml` change — confirm `uv lock --check` passes against the unchanged pyproject.
- ❓ q: PR `mergeable_state` is `dirty` — branch has conflicts with `master`.

abhisheksharma001 (commented):
## Review

**High**
1. **`website/components/RunListPanel.tsx:162`: rename input loses focus after every keystroke.** `Row` is defined inside the panel's render and used as `<Row/>`, so React sees a new component type each render and remounts the row. Each keystroke updates state, re-renders, and throws away focus. Move `Row` out of the render function or render the rows inline, as the old code did.
2. **`website/components/RunListPanel.tsx:274` + `website/lib/run-filters.ts:78`: clicking a hotspot cell filters the wrong runs.** The `origin` and `node` filters read `run.origins` / `run.finding_nodes`, but the Python `/api/runs` summary (`_list_runs`) never sends those fields. The origin filter falls back to `first_failure_step === X`, so it often shows "No runs match this filter". The node filter matches any run whose graph merely contains that node. Add both fields to the run summary, or filter on the server side.

**Medium**
3. **`website/components/HotspotMatrix.tsx:55`: `/api/hotspots` doesn't exist on the server.** `aggregate_hotspots` in `src/argus/hotspots.py` is never wired to a route. The matrix always falls back to fetching up to 40 full run records. That caps coverage at 40 runs instead of 500, and the fetches re-run on every run-list change, which is every 6s poll in live mode. Add the route.
4. **`website/components/run-detail/ExecutionGraph.tsx:123`: `dagLayers` counts edges from sources that aren't nodes (e.g. `__start__`).** Every entry node then starts with in-degree 1 and only `names[0]` seeds the layout. Parallel entry points collapse into the wrong column, and graphs whose entry node isn't `names[0]` pile into one column. Skip edges whose source isn't in `names`, as the old code did.
5. **`website/components/run-detail/ExecutionGraph.tsx:74`: `mapStatus` has no `interrupted` case,** so it falls through to `default: 'pass'`. Human-in-the-loop pauses render green, and the `running` style can never be reached. Map `interrupted` to `running`, as before.
6. **`website/app/compare/components/StructuredDiff.tsx:62`: `` `${highlightColor}08` `` produces `var(--tool)08`, which is invalid CSS,** so the changed-row highlight is dropped. Use `color-mix(in srgb, var(--tool) 3%, transparent)` or a dedicated token.
7. **`website/lib/workspace.tsx:88`: `fetchRuns` keeps the old list when `run_id` + `overall_status` are unchanged.** Renames from another tab and in-progress `step_count` / `duration_ms` changes never show up. Compare more fields, or just replace the list.

**Low**
8. **`website/components/HotspotMatrix.tsx:63`: no `cancelled` re-check after `await fromDetails()`.** A stale response can overwrite newer data.
9. **`website/app/approvals/ApprovalRows.tsx:62`: the same `${color}10` / `${color}30` pattern on `var(...)` colours** makes the strategy badge background and border invalid. `RunTable.tsx:188` has the same pattern, but that component is currently unused.
10. **`website/components/run-detail/PipelineTab.tsx:37`: the `Tree` view has the same `__start__` in-degree bug as #4.** Only `names[0]` becomes a root.
11. **`src/argus/hotspots.py:56`: `_matches_tag` can raise a TypeError.** `value in tags.get(key, [])` fails when the value is `None` or a number, and the no-colon branch does substring matching against `str(tags)`. Nothing calls it today.

Visual review and graph interaction (the unchecked boxes in the PR description) are still worth doing before merging.

PR #74, opened 2026-09-09
fix(patcher): don't corrupt subgraph nodes lacking a wrappable func

abhisheksharma001:
## Problem

`patch_graph` crashes LangGraph compile for graphs containing **subgraph nodes** — `StateNodeSpec` entries whose `.runnable` is a compiled graph (`CompiledStateGraph`) rather than a `RunnableCallable`. These have no `.func` to wrap, so the code falls through to the legacy branch and replaces the whole spec with a plain function:

```
AttributeError: 'function' object has no attribute 'ends'
  langgraph/graph/state.py in validate
```

Repro: attach `ArgusWatcher` to a compiled `langgraph-swarm` graph (`create_swarm(...)` puts each agent's compiled graph in `graph.nodes`) or a `langgraph-reflection` graph — attach recompiles via `_attach_compiled` and validate blows up. Same corruption path exists for the uncompiled `attach(builder)` flow.

## Fix

Only take the legacy replace branch for true plain callables (no `.runnable` attribute). Subgraph nodes are left unmonitored instead of corrupting `graph.nodes` — a subgraph's inner nodes remain individually watchable when the subgraph itself was compiled with ARGUS attached.

## Tests

`tests/test_patcher_graph.py` — regression coverage for all three node shapes:
- plain callable → wrapped (legacy path preserved)
- `StateNodeSpec` with `.func` → wrapped in place, spec object preserved
- `StateNodeSpec` wrapping a compiled graph → skipped, spec untouched

All 3 pass; verified end-to-end against a live `langgraph-swarm` 0.1.0 graph with `ArgusWatcher().attach()` persisting runs.


VaradDurge:
Thanks for this — the subgraph-node guard is a real gap in the current patcher.

We are pivoting capture **off** engine wrapping as the default path (patcher, `StateGraph.compile` wrapping, Pregel `invoke`/`stream`/`batch`, live node wrappers). The founder is replacing that core with a fat-trace recorder that feeds the existing inspector / `argus check`. Please do not invest further in wrapping LangGraph at the engine root.

This PR is **on hold**, not rejected. It extends the old core we are replacing, so we will not merge it as-is. Once the fat-trace recorder exists, we can retarget this work if the issue still applies, or close it if wrapping subgraphs is no longer relevant.

Appreciate the careful fix.

VaradDurge (approved):

PR #75, opened 2026-09-09
fix: preserve store/config kwargs in watcher node wrapper

abhisheksharma001:
## Problem

`ArgusWatcher.attach()` on an already-compiled graph recompiles the underlying builder (`_attach_compiled`) but only carried over `checkpointer` / `interrupt_before` / `interrupt_after`. A graph compiled with `compile(store=...)` (or `cache=...`) silently lost it: langgraph's runtime kwarg injection then handed nodes `store=None` / `cache=None`.

Repro (before the fix, `store` inside the node is `None`; after, it's the `InMemoryStore`):

```python
from typing import TypedDict
from langgraph.graph import END, START, StateGraph
from langgraph.store.base import BaseStore
from langgraph.store.memory import InMemoryStore
from argus import ArgusWatcher

class S(TypedDict):
    n: int

def node(state: S, *, store: BaseStore) -> dict:
    print(type(store).__name__)   # NoneType before fix, InMemoryStore after
    return {"n": state["n"] + 1}

g = StateGraph(S)
g.add_node("node", node)
g.add_edge(START, "node")
g.add_edge("node", END)

app = g.compile(store=InMemoryStore())
ArgusWatcher(investigate=False, semantic_judge=False).attach(app)
app.invoke({"n": 0})
```

## Fix

Carry `store` and `cache` over to the recompile alongside the existing kwargs in `_attach_compiled`.

## Tests

New `tests/test_watcher_store_injection.py` (2 tests): compiled-app `attach()` path and uncompiled-builder constructor path. Both fail before the fix (compiled path), pass after. Unrelated to PR #74 (patcher subgraph fix) — separate branch off master.


VaradDurge:
Thanks for catching the dropped `store`/`config` kwargs — that is a genuine wrapper bug.

We are pivoting capture **off** engine wrapping as the default path (patcher, watcher node wrappers, Pregel runtime wrapping). The founder is replacing that core with a fat-trace recorder that feeds the existing inspector / `argus check`. Please do not invest further in wrapping LangGraph at the engine root.

This PR is **on hold**, not rejected. It extends the old core we are replacing, so we will not merge it as-is. Once the fat-trace recorder exists, we can retarget this if kwargs still matter on a wrapper, or close it if that path is gone.

Appreciate the work here.

VaradDurge (approved):

PR #84, opened 2026-09-10
fix(ledger): honour reducers when computing state_after

### LINKED ISSUE #80: [Fat-Traces] Ledger state_after is reducer-blind for Annotated[..., operator.add] fields
Branch: `pivot/fat-traces`. Fully specified — good first issue.

## The bug

`build_ledger` (`src/argus/ledger.py`) computes each row's `state_after` with a plain dict overlay: last write wins. For a field declared with a reducer that is wrong.

Measured on a 3-node graph with `docs: Annotated[list, operator.add]`, where node `b` returns `{"docs": []}`:

```
REAL final docs       : ['seed']     # what LangGraph actually holds
LEDGER state_after[b] : []           # what the notebook says
```

The notebook is meant to be the source of truth for re-score. Right now it lies about reduced fields.

## Not currently a false positive — but one bad tightening away

Grading is unaffected today: `session.py:760-771` merges with the real reducer callables, and `contextual._lacks` only treats `None` as missing, so `[]` does not fire. Tighten `_lacks` to also treat `[]` as dropped — a plausible next step — and this misfires immediately.

## Why it was not fixed live-only

Reducers are callables. They do not survive the run file. Populating them only when the recorder is live would make a reloaded notebook differ from the live one, breaking the property spike 2 exists to guarantee.

**Hard constraint:** `tests/test_ledger.py::test_ledger_from_live_steps_matches_the_reloaded_one` must still pass. Live and reloaded notebooks must not diverge.

## Two candidate fixes

- Persist reducer identity (import path or a known-reducer enum) in the run file and re-resolve on load.
- Rebuild `state_after` from each successor's recorded `input_state` — that is ground truth straight from the framework, no reducer emulation. Needs the edge map to pick the right successor under fan-out, and the last step still needs the overlay.

The full note, with the repro, is in the `ponytail:` comment at the top of `build_ledger`.

## Done when

One test that fails on `main` and passes with the fix, asserting `state_after` for an `operator.add` field matches what the graph really returned.
---

### Where to work

**Base this on `pivot/fat-traces`, not `master`.** GitHub will pre-fill `master` as the PR base — change it.

```bash
git fetch origin
git checkout -b <your-branch> origin/pivot/fat-traces
pip install -e ".[dev]"
```

Start here: [`docs/PIVOT-BRANCH.md`](../blob/pivot/fat-traces/docs/PIVOT-BRANCH.md).

Do not run the whole suite while iterating — `tests/test_async_judge_e2e.py` can block on a live network call. For the pivot:

```bash
pytest tests/test_recorder.py tests/test_ledger.py tests/test_contextual.py tests/test_judge_last.py -v
python demo/fat_trace/demo_graph.py && argus check    # expect exit 1
```

HarshRajSinghania:
## Summary

`build_ledger` used a plain dict overlay for `state_after` (last write wins). For fields declared as `Annotated[list, operator.add]`, a node returning `{"docs": []}` therefore overwrote the accumulated list even though LangGraph kept it.

This rebuilds `state_after` from the next recorded step's `input_state` when one exists. That value is the framework's post-reducer merge, so reduced fields stay correct without persisting reducer callables. Live and reloaded notebooks stay identical. The last step still uses the overlay (no successor).

Fixes #80.

## Motivation

Issue #80: the notebook is the source of truth for re-score, but it lied about reduced fields. Grading is unaffected today, but tightening `_lacks` to treat `[]` as missing would immediately misfire.

## Implementation

Chosen the second candidate from the issue: rebuild `state_after` from each successor's recorded `input_state` rather than persist reducer identity. That keeps the run file unchanged and satisfies `test_ledger_from_live_steps_matches_the_reloaded_one`.

Limitation: fan-out without a recorded next event still falls back to last-write-wins for that row. Linear graphs and sequential recordings are covered.

## Testing

```
PYTHONPATH=src pytest tests/test_ledger.py tests/test_recorder.py tests/test_contextual.py tests/test_judge_last.py -v
```

- `tests/test_ledger.py`: 3 passed (including new `test_state_after_honours_operator_add_reducer`)
- `tests/test_contextual.py`: 4 passed
- `tests/test_judge_last.py`: 5 passed
- `tests/test_recorder.py`: 9 passed, 1 failed (`test_replay_refuses_a_trace_run_loudly`) with `ModuleNotFoundError: typer` — environment gap, unrelated to this change

### DIFF
--- a/src/argus/ledger.py
+++ b/src/argus/ledger.py
@@ -50,25 +50,23 @@ def build_ledger(
     Retried and skipped events are kept: later rows are the evidence that a
     downstream node was the victim rather than the origin.
     """
-    # ponytail: plain dict overlay for the running state — last write wins.
-    # Measured limitation: for a field declared `Annotated[list, operator.add]`,
-    # a node returning `{"docs": []}` leaves `state_after["docs"] == []` here
-    # while LangGraph really keeps the accumulated list. Grading is unaffected
-    # (session.py merges with the real reducers, and `_lacks` in contextual.py
-    # only treats None as missing), but the notebook's running state is wrong
-    # for reduced fields.
-    # Not fixed live-only on purpose: reducers are callables and do not survive
-    # the run file, so a live-only fix would make the reloaded notebook differ
-    # from the live one and break re-score. Fixing it properly means persisting
-    # reducer identity, or rebuilding state_after from each successor's recorded
-    # input_state (needs the edge map for fan-out).
+    # Running state is rebuilt from each successor's recorded input_state when
+    # one exists. That value is the framework's post-reducer merge, so reduced
+    # fields (e.g. Annotated[list, operator.add]) stay correct without storing
+    # reducer callables — live and reloaded notebooks stay identical.
+    # The last step has no successor, so it still uses a last-write-wins overlay.
     running: dict[str, Any] = dict(initial_state or {})
     rows: list[LedgerRow] = []
+    events = list(steps)
 
-    for event in steps:
+    for i, event in enumerate(events):
         update = event.output_dict
         if update:
             running = {**running, **update}
+        if i + 1 < len(events):
+            nxt = getattr(events[i + 1], "input_state", None)
+            if nxt:
+                running = dict(nxt)
         rows.append(
             LedgerRow(
                 step_index=event.step_index,
--- a/tests/test_ledger.py
+++ b/tests/test_ledger.py
@@ -5,7 +5,8 @@ Invoke once, load the saved run, build the ledger from *that*, and grade it.
 
 from __future__ import annotations
 
-from typing import TypedDict
+import operator
+from typing import Annotated, TypedDict
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


VaradDurge:
Holding until the fat-trace pivot lands.

This changes `src/argus/ledger.py` (`state_after` / reducer merge), which is the new detection notebook. We are not merging detection-core PRs while that path is still moving. Not rejected — retarget against `pivot/fat-traces` once the recorder + ledger contract is stable, or reopen if this still applies then.

abhisheksharma001:
## Review

- `src/argus/ledger.py:66`: 🟡 risk fan-out regression mischaracterized as limitation. `input_state` is snapshotted at `on_chain_start`, so for parallel nodes in the same super-step every sibling gets the pre-super-step snapshot — row for node `b` adopts sibling `c`'s `input_state` and silently drops `b`'s own update from `state_after` (strictly worse than the old overlay, which at least kept it). Only last-finishing node of a parallel batch gets correct post-reducer state. Fix: adopt `nxt.input_state` only when `nxt` is a verified sequential successor (edge map from `session.set_edges`, or `step_index` ordering vs recorded start order), otherwise keep the overlay.
- `src/argus/ledger.py:68`: 🔵 nit `if nxt:` skips a legitimately empty `{}` input_state and keeps the stale overlay; use `if nxt is not None:`.
- `src/argus/ledger.py:57`: 🔵 nit last row still last-write-wins for reduced fields — if `_lacks` is ever tightened as #80 motivates, the final row will still misfire; consider at least marking it.
- `tests/test_ledger.py:133`: 🔵 nit test reaches into private `recorder.session._events` / `_initial_state`; a public accessor on `ArgusSession` would keep the test decoupled from internals.

Core approach is sound for linear graphs: successor `input_state` is the framework's post-reducer merge, and the live-vs-reloaded equality assertion covers the persistence round-trip.

abhisheksharma001 (commented):
## Review

The fix is correct for linear graphs. In other graph shapes it regresses compared with the old overlay.

- **`src/argus/ledger.py:66-69` (MAJOR): fan-out regression.** Steps in the same superstep all record the same pre-superstep `input_state`, and events are stored in completion order. With parallel siblings `b` and `c`, `b.state_after` becomes `c.input_state`, so `b`'s own update disappears. Only take `events[i+1].input_state` when the successor is a direct edge target and the only step in its superstep (this needs the edge map or a superstep/parent id on `NodeEvent`). Otherwise keep the overlay.
- **`src/argus/ledger.py:69` (MAJOR): `running = dict(nxt)` replaces the whole running state.** When the next node has an `input=` schema, private channels, or is reached through `Send`, its input is narrower than the full state, so keys vanish and `contextual._lacks` reports false critical `missing_field` findings. Merge instead (`running = {**running, **nxt}`) and skip successors reached through `Send`.
- **`src/argus/ledger.py:66` (MINOR):** across several `invoke()` calls in one session without `begin_new_run`, the last step of one call inherits the next call's input as its `state_after`. Only look ahead within the same run.
- **`tests/test_ledger.py` (MINOR):** only a linear 3-node graph is tested. Please add fan-out (two parallel writers), input-schema and `Send` cases.

PR #98, opened 2026-09-14
S-2: extract the grading chain into grading.py

abhisheksharma001:
**Stacked on `s1-langsmith-fixture`** — new here: `00bf62a` (code) and `05eeb42` (PRD note). Review the earlier PRs first.

## What changed
- `src/argus/grading.py` (new): `new_session(node_names, edge_map, conditional_sources, reducer_fields, *, judge, validators, strict, max_field_size)` and `finish(session, consumers, unfinished=())`, plus `IncompleteTraceError`, `_placeholder_node`, `_blame_origins`, `_refuse` — moved verbatim from `recorder.py`. Imports nothing from `langgraph` / `langchain_core`.
- `src/argus/recorder.py`: keeps `_topology`, the subgraph-parent skip, and its unfinished-step bookkeeping; `_new_session` / `_finish` are now thin wrappers (tests call them). `IncompleteTraceError` still importable from `argus.recorder`.
- `CLAUDE.md`: one file-table row for `grading.py`.

## Proved
- Full suite unchanged: 1018 passed, 4 skipped, 2 xfailed. No test edited.
- `argus.grading` imports with `langgraph` / `langchain_core` blocked.
- Break test: removing the empty-trace refuse from `grading.py` fails `test_recorder.py::test_an_empty_trace_refuses_to_grade` and `test_silent_failure_matrix.py::test_a_skinny_trace_is_refused_rather_than_graded_clean`; restored, green.

## Left out
No behaviour or detection-rule change. PRD signature said `reducer_kinds`; it must be `reducer_fields` (the session merges with the callables) — corrected in the PRD.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LSb2PaZf8y3PWrKeUzkBx3

PR #97, opened 2026-09-14
S-1: offline LangSmith-shaped fixture from the demo graph

abhisheksharma001:
**Stacked on `prd-abhishek`** — only the top commit (`61b70fe`) is new here; review that PR first.

## What changed
- `scripts/make_langsmith_fixture.py`: runs `demo/fat_trace/demo_graph.py` through the real `LangChainTracer` with a stub client (`create_run` / `update_run` / `flush` record payloads, send nothing). Writes rows in export shape: every field present, `null` when never sent; `extra.runtime` stripped.
- `tests/fixtures/langsmith/demo_graph.jsonl`: 4 runs — `LangGraph` root (full merged state), `search` (`docs`), `summarize` (`{}`), `answer`.

## Proved
- The PRD S-1 check passes (`ok 4`): `summarize` outputs `{}`/absent, `search` has `docs`, one root with null `parent_run_id` and non-empty outputs.
- ruff clean; full suite 1018 passed, 4 skipped, 2 xfailed.
- The probe is what found BUG-1.

## Left out
No network, no `langsmith.Client`, no `src/` change. No break test — data only, no guard. Ids and timestamps change per regeneration; tests must not pin them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LSb2PaZf8y3PWrKeUzkBx3

PR #99, opened 2026-09-14
S-3: argus ingest langsmith grades an exported run with no app

abhisheksharma001:
**Stacked on `s2-grading-module` (#98)** — new here: the S-3 code commit and its PRD note. Review #96 → #97 → #98 first.

## What changed
- `src/argus/ingest/langsmith.py` (new): reads a LangSmith JSONL export, merges rows by `id`, keeps `chain` runs with `extra.metadata.langgraph_node` and a `graph:step:N` tag, drops subgraph parent rows, orders by `langgraph_step` then `dotted_order`, and replays each into a `grading.new_session`:
  - `outputs` absent or `{}` with no error → the `{}` update (BUG-1); non-empty dict → `capture_output`; `error` → a `TracedError` whose `str()` is the error text verbatim, update `None`.
  - Successors from observed step order (each node → nodes at the next step), so `empty_output` fires; real edges are S-5.
  - `grading.finish` saves the run; `argus check last` gates it.
- Judge off for ingest — grading a file spends nothing.
- Refuses to save while logged in to ARGUS cloud unless `--allow-cloud` (`save_run` uploads).
- `src/argus/cli/cmd_ingest.py` + `argus ingest langsmith <file>` in `main.py`; exit 2 when a file cannot be graded.
- No `langgraph` / `langchain_core` import.

## Proved
- `argus ingest langsmith tests/fixtures/langsmith/demo_graph.jsonl` → `[argus] run … silent_failure on summarize`; `argus check last` exits 1, `first_failure_step == "summarize"`.
- `tests/test_ingest_langsmith.py`: 4 tests (blame from file, cloud refusal, subgraph parent skip, import with langgraph/langchain blocked). Full suite 1022 passed, 4 skipped, 2 xfailed. No LLM call, no network.
- Break tests: removing the cloud guard fails exactly the refusal test; `{}` edges fails exactly the blame test.

## Left out
Tool child runs (S-4), real edges (S-5), skinny-trace refusal on hidden outputs (S-6), `--consumers` (S-7). PRD verify used `python -m argus.cli.main`, which silently does nothing (no `__main__` guard) — corrected to call `app()`.


VaradDurge:
Rebased and force-pushed — third in the stack, same cause as #97 and #98 (#96–#98 all landed as squash merges, so this branch was carrying their commits as duplicates). Replayed just your two:

```
git rebase --onto origin/pivot/fat-traces 05eeb42 s3-ingest-langsmith
```

New head is `83daa08`; net diff is now only S-3's six files.

**Verified end to end before merging:**
- `argus ingest langsmith demo_graph.jsonl` in a clean temp dir → `silent_failure on summarize`, `empty_output on _output`. Run record: `first_failure_step == "summarize"`, steps `search=pass, summarize=fail, answer=pass`, and `argus check last` exits 1. Blame lands on the silent node, not the downstream reader, and the two healthy nodes stay clean — the no-false-positive half matters as much as the catch.
- Both break tests reproduce, and each fails *only* its own test: removing the cloud guard fails `test_logged_in_refuses_to_save_without_allow_cloud`; `_step_order_edges` → `{}` fails `test_the_silent_node_is_blamed_from_the_file_alone`. Load-bearing guards, not decorative ones.
- `argus.ingest.langsmith` and `argus.grading` both import with `langgraph` and `langchain_core` blocked via a `meta_path` hook.
- Full suite 1022 passed / 4 skipped / 2 xfailed; ruff clean on `src/` and `tests/`. CI green on 3.9/3.11/3.12.

Two things I checked specifically because they'd fail quietly rather than loudly:

1. **`capture_state` ordering.** It latches `_initial_state` only on the first non-empty snapshot, so calling it for the root's inputs *before* the per-step loop is what makes the ledger's initial state right. Correct as written — worth a comment someday, since reordering those two lines would corrupt the ledger without failing a test.
2. **Subgraph parent detection.** Walking each selected run's ancestry and marking any ancestor that is itself selected is the right shape — it matches the recorder's skip and won't mistake the `LangGraph` root for a subgraph parent.

Good catch on `python -m argus.cli.main` being a silent no-op with no `__main__` guard, and on fixing the PRD rather than writing a verify step that appears to pass while doing nothing.

**To sync locally:**

```
git fetch origin && git checkout s3-ingest-langsmith
git reset --hard origin/s3-ingest-langsmith
```

Check `git log origin/s3-ingest-langsmith..s3-ingest-langsmith` first if you have unpushed work. S-1 through S-3 are now all on `pivot/fat-traces`, so S-4 can branch straight off it — no more stacking needed.

PR #100, opened 2026-09-14
fix(recorder): preserve all contextual reasons

### LINKED ISSUE #94: [Fat-Traces] _blame_origins overwrites inspection.message when one node misses several fields
## Good first issue — small, self-contained, has an obvious test.

`_blame_origins` copies each contextual finding onto the origin step (`recorder.py:489-509`):

```python
for finding in findings:
    ...
    insp.missing_fields.append(finding.field_path)
    insp.is_silent_failure = True
    insp.severity = "critical"
    insp.message = finding.reason     # <-- last one wins
```

When one node is the origin for two or more missing fields, `missing_fields` correctly accumulates both — but `message` is overwritten each time, so only the last field's sentence survives. `argus check` prints `reasons[0]` and the CLI/UI surface `inspection.message`, so the contributor reading the output sees one field explained and the other listed with no reason.

It also clobbers any message an earlier layer (structural inspection, tool scan) already wrote on that same step, replacing a deterministic explanation with a contextual one for no stated reason.

## Expected

The message reflects every reason attached to the step — join them, or keep the existing message and append. Don't drop information that the findings list already has.

## Where to fix

`recorder.py:489-509`. Note `RunRecord.findings` (via `findings.collect_findings`) already gets one `Finding` per signal with the full sentence, so the fix is only about the summary text on the inspection — keep it consistent with what the findings list says.

Test: one node that is the origin for two declared consumer fields; assert both field names appear in the reported reason.

MinhaKim02:
## What & why

Closes #94.

`ArgusRecorder._blame_origins` previously replaced `inspection.message` for every contextual finding, so only the last missing-field reason survived. This change:

- replaces the default success message with the first contextual reason;
- appends subsequent contextual reasons without duplicating them;
- preserves any structural or tool message already attached to the origin step;
- adds an integration regression test with two declared consumer fields blamed on the same origin.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / internal
- [ ] Docs
- [ ] Other:

## Checklist

- [x] I have joined the [ARGUS Discord](https://discord.gg/67XTFTDSgd) (updates, questions, PR discussion).
- [ ] I have signed the [CLA](../CLA.md) (the bot will prompt on first PR).
- [x] My change targets the **open-source core** only — I have **not** modified
      `cloud/` or `supabase/` (proprietary; maintainer-only).
- [x] Tests added/updated and `pytest` passes locally.
- [x] `ruff check .` passes.
- [x] Docs updated if behavior or the public API changed (`CLAUDE.md`, relevant docs).
- [ ] PR has a label (`bugfix`, `enhancement`, `refactor`, `investigation`, `documentation`, `dead-code`).

## Notes for reviewers

Local verification on the latest `pivot/fat-traces`:

- `pytest tests/ -q`: 1017 passed, 6 skipped, 2 xfailed
- `pytest tests/test_argus_ci_gate.py --argus --tb=short -q`: 1 passed
- `ruff check .`: passed


MinhaKim02:

VaradDurge:
Thanks — real bug, correctly diagnosed. `insp.message = finding.reason` kept only the last reason and clobbered whatever the structural or tool layer had already authored on that step.

I rebased and force-pushed (new head `d952930`, your authorship preserved). This one had a genuine conflict rather than a stale-branch one: #98 landed in the meantime and moved `_blame_origins` out of `recorder.py` into the new `src/argus/grading.py`. So your fix now lives there, unchanged in behaviour:

```python
if insp.message == "All checks passed":
    insp.message = finding.reason
elif finding.reason not in insp.message:
    insp.message = f"{insp.message}; {finding.reason}"
```

That's a better home for it than where you found it — `grading.py` is the engine-free chain, so the fix now also covers `argus ingest langsmith` (#99), which reaches `_blame_origins` through the same `finish()`. I moved your `CLAUDE.md` sentence onto the `grading.py` row to match, and kept the `docs/PIVOT-BRANCH.md` paragraph as written.

**Verified:**
- Your new test genuinely catches the bug: reverting to the one-line overwrite fails `test_contextual_reasons_accumulate_on_the_same_origin`; with the fix, it passes. A regression test that fails for the right reason is the part people skip, so thanks for including it.
- Both detection matrices: 54 passed. Full suite 1023 passed / 4 skipped / 2 xfailed. ruff clean. CI green on 3.9/3.11/3.12.

Two notes, neither blocking:

1. Your third bullet — "preserves any structural or tool message already attached" — is the behaviour I'd most want pinned, and it's the one case the new test doesn't reach. It only exercises the `elif` via a *second contextual* reason, not via a pre-existing structural message. The code handles it, but nothing would catch a regression there. If you want a follow-up, that's a good small one.
2. `"All checks passed"` as a literal is slightly fragile, but it's already the de-facto sentinel compared this exact way in `findings.py:83` and `cmd_open_ui.py:1688`, so matching local convention was the right call over inventing a constant.

Merging. Your branch was force-pushed, so to sync locally:

```
git fetch origin && git reset --hard origin/fix/blame-origins-message
```

(Also approved the CI workflow run — first-time fork contributions need a maintainer to release them, which is why your checks sat idle rather than failing.)

Issue #86, opened 2026-09-14
[Fat-Traces] Recorded tool errors never reach the graders — a failing tool grades clean


VaradDurge:
## What happens

`ArgusRecorder` captures real tool I/O through `on_tool_start` / `on_tool_end` / `on_tool_error` and files it on the step (`recorder.py:408` → `NodeEvent.tool_calls`). **Nothing ever grades it.** The tool-failure scan (`inspector.inspect_tool_outputs`) only reads the node's *output dict* — it has no idea `event.tool_calls` exists (`grep -rn inspect_tool_outputs src/argus/` → only `inspector.py` and one call site inside `inspect_transition`).

So a tool that raised, returned `{"error": ...}`, or came back empty is recorded faithfully and then graded **clean**, as long as the node swallowed it and returned a plausible-looking update. That is exactly the silent failure ARGUS exists to catch, and `CLAUDE.md` claims the tool-failure scan is "unchanged" on the pivot path — it is not, it is simply unreached.

## Repro (verified, langgraph 0.6.x)

```python
@tool
def search(q: str) -> str:
    """search"""
    raise RuntimeError("upstream 500")

def n1(s, config=None):
    try:
        search.invoke({"q": "x"}, config=config)
    except Exception:
        pass          # the classic swallow
    return {"a": "fallback"}
```

Graph `n1 -> n2`, run under `ArgusRecorder().attach(app)`:

```
n1 pass tools= [{'name': 'search', 'input': "{'q': 'x'}", 'output': None,
                 'error': "RuntimeError('upstream 500')"}]
20260914-103451-f41db7  clean
findings: []
```

The evidence is in the trace. The verdict ignores it.

## Expected

A recorded tool with `error` set, or an empty/error-shaped `output`, produces a `tool_failure` finding blamed on the node that called it, and fails `argus check`.

## Where to fix

`session.on_node_end` already runs the inspection; the tools are attached *after* it returns (`recorder.py:405-409`). Either pass `tools` into `on_node_end` so the scan sees them, or add a ledger-level pass (`build_ledger` rows already carry `tools`) that emits findings the way `contextual.contextual_findings` does. The ledger route is probably the smaller diff and matches the pivot's "grade the notebook" shape.

Guard with a case in `tests/test_silent_failure_matrix.py` (the tool-fetcher pipeline is already there) plus a false-positive case: a tool that succeeded must stay clean.

VaradDurge:
[...]

Issue #89, opened 2026-09-14
[Fat-Traces] A subgraph can write nothing the parent graph sees, and the ledger carries its inner-only keys


VaradDurge:
## What happens

`_topology()` reads `get_graph(xray=True)` so nodes *inside* a subgraph are registered — that part works. But x-ray **replaces** the parent node's edges with the inner nodes' edges, so the parent step ends up with no successors at all.

Verified on a `parent(child -> after)` graph where `child` is a compiled subgraph:

```
NODES ['__start__', 'after', '__end__', 'child:inner1', 'child:inner2']
EDGE __start__ -> child:inner1
EDGE child:inner2 -> after
EDGE child:inner1 -> child:inner2
EDGE after -> __end__

_topology(app) ->
  names       ['after', 'inner1', 'inner2', 'child']
  edge_map    {'inner2': ['after'], 'inner1': ['inner2']}
```

`child` is in `names` (the un-x-rayed union puts it back) and the callbacks do report a `child` step — but `edge_map` has no entry for it. `session.on_node_end` passes `has_successors=bool(self.graph_edge_map.get(node_name))`, so the `empty_output` rule is switched off for that step, even though `after` is plainly waiting on it.

Same shape applies to any graph where a node is a compiled subgraph, which is the standard way to compose agents.

## Expected

The subgraph parent step keeps its own successors. Nodes inside it keep theirs. Both are graded.

## Where to fix

`recorder.py:125-156` (`_topology`). The un-x-rayed graph already has the right answer for the parent (`child -> after`); today it is only mined for node *names*. Union its edges into `edge_map` too rather than taking the x-rayed edge list alone — and check the two do not contradict each other for a node that exists in both.

Guard: extend the subgraph pipeline in `tests/test_silent_failure_matrix.py` with a case where the **subgraph itself** contributes nothing and a downstream node waits on it. The existing subgraph test covers inner nodes, which is why this stayed hidden.

VaradDurge:
Fixed on `pivot/fat-traces` — but the diagnosis in this issue turned out to be stale, and what was actually broken was worse. Recording that here so the reasoning isn't lost.

## The stated mechanism no longer exists

The issue says `child` lands in `names`, the callbacks report a `child` step, and `edge_map` has no entry for it, so `empty_output` is switched off for that step. Reproducing the exact `parent(child -> after)` graph on current `pivot/fat-traces`:

```
[...]

Issue #87, opened 2026-09-14
[Fat-Traces] Graph invoked inside an outer chain records nothing — and says nothing


VaradDurge:
## What happens

`on_chain_start` attributes a callback to a run like this (`recorder.py:294`):

```python
root = run_id if parent_run_id is None else self._root_of.get(parent_run_id)
if root is None:
    return  # a callback we cannot attribute to any run we started
```

A parentless chain is treated as the run boundary. But when the compiled graph is one step of a larger LCEL chain — `prompt | app`, `RunnableLambda(...) | app`, a graph used as a tool, a LangServe route — the graph's own chain **has** a parent, and that parent is a chain the recorder never saw. `root` is `None`, we `return`, and every subsequent callback for that run also fails to resolve.

Result: no session, no run file, no verdict — and **no error**. `IncompleteTraceError` never fires because `_finish` is never reached. This is the "no findings, so it passed" case the brief bans, arriving through the front door.

## Repro (verified)

```python
g.add_node("n1", lambda s: {})      # silent no-op, has a successor
g.add_node("n2", lambda s: {"a": "x"})
rec = ArgusRecorder()
app = rec.attach(g.compile())
chain = RunnableLambda(lambda x: x) | app
chain.invoke({})
# OUT: {'a': 'x'}
# rec.run_ids: []   rec.session: None
```

Invoking `app` directly on the same graph correctly reports `silent_failure on n1`. Composing it into anything at all makes ARGUS silently stop working.

## Expected

Either:

- recognise the graph run itself rather than relying on "parentless" — the graph chain is identifiable from `metadata` / `tags` (e.g. `langgraph_step` absent + a `LangGraph`-ish `serialized["name"]`, or the first chain that later spawns `langgraph_node` children), or
- keep the parentless heuristic but **refuse loudly**: if node-level callbacks (`langgraph_node` in metadata) arrive with no resolvable root, raise `IncompleteTraceError` instead of dropping them.

The second is the lazy floor and should land regardless of the first — a recorder that silently records nothing is worse than one that crashes.

## Where to fix

`recorder.py:278-337` (`on_chain_start`), with the refusal path next to `_refuse` (`recorder.py:511`).

VaradDurge:
Fixed on `pivot/fat-traces` (`6cbc48b`). The symptom you described is exactly right; the cause turned out to be one layer lower, so recording that here.

## The stated cause is not what happens

[...]

PR #101, opened 2026-09-14
S-4: tool child runs land on the ledger row on ingest

abhisheksharma001:
**Stacked on #99 (S-3)** → #98 → #97 → #96. Review in order; this diff's own commits are the top three.

> #96–#99 are now squash-merged, so this branch conflicts; a rebased push onto 87d7706 is pending.

## What changed
- `src/argus/ingest/langsmith.py`: `tool_calls_by_step` — each `run_type == "tool"` run is attached to its nearest node-step ancestor (walking `parent_run_id`), in the dict shape `ArgusRecorder.on_tool_start` builds (`name`, `input`, `output`, `error`), and passed as `on_node_end(..., tool_calls=)`. Tools go in with the step, never after it — graders run inside `on_node_end` (#86).
- LangSmith stores a tool result as `outputs = {"output": <result>}`; the recorder hears the bare result, so ingest unwraps it.
- `scripts/make_langsmith_fixture.py --tool`: a second graph whose `fetch` node calls a tool returning `{"status": 500, "body": "upstream down"}`, swallows it, and returns a normal update. Generated offline with a stub client → `tests/fixtures/langsmith/tool_graph.jsonl`.
- No `inspector.py` rule changes: the finding comes from the existing `inspect_tool_calls`.

## Proved
- Ingesting `tool_graph.jsonl` → `argus check last` exits 1: `silent_failure on fetch`, critical `error_response on fetch_docs.status`.
- Break proofs: `tool_calls=[]` fails exactly `test_a_swallowed_tool_500_fails_the_node_that_called_the_tool`; removing the unwrap fails exactly `test_a_tool_result_is_unwrapped_to_what_the_recorder_hears` (detection alone didn't catch it — the nested scan still finds `output.status`).
- Full suite: 1024 passed (1025 after rebasing locally on 87d7706), 4 skipped, 2 xfailed. No LLM call, no network.

## Left out
- Real graph edges (S-5); tool *errors* from a real export are passed through as LangSmith's error string, untested against a live export.


VaradDurge:
Rebased and force-pushed — #96–#99 are squash-merged as you noted, and the base has moved further since (#100, plus #89 and #87 fixes). Replayed your three commits onto current `pivot/fat-traces`:

```
git rebase --onto origin/pivot/fat-traces bae5038 s4-tool-child-runs
```

New head is `8bf6842`. Net diff is now only S-4's five files. Nothing of yours changed in the rebase — `grading.new_session` gained an optional `state_keys` argument in the meantime, and ingest correctly does not pass it (a trace file has no schema to read).

**Verified:**
- `argus ingest langsmith tool_graph.jsonl` → `silent_failure on fetch`, critical `error_response on fetch_docs.status`, `argus check last` exits 1, and the tool lands on the step with the payload intact. `answer` stays clean.
- Both break proofs reproduce and each fails *only* its own test: `tool_calls=[]` fails `test_a_swallowed_tool_500_fails_the_node_that_called_the_tool`; removing the unwrap fails `test_a_tool_result_is_unwrapped_to_what_the_recorder_hears`.
- Full suite 1041 passed / 4 skipped / 2 xfailed; both matrices 68 passed; ruff clean; CI green on 3.9/3.11/3.12.

Attaching a tool to its nearest *step* ancestor rather than its direct parent is the right call — it survives the inner runnables LangChain puts between a node and its tool, and it's what makes this work on a react agent rather than only on the fixture.

**One change I made on top** (`8bf6842`), because it was a silent-loss path:

```python
"output": outputs.get("output", outputs) if isinstance(outputs, dict) else outputs
```

`outputs["output"]` is LangSmith's shape but not a guarantee. Reading the key blind returns `None` for an export that stores the payload directly — I confirmed it: force `outputs = {"status": 500, "body": "upstream down"}` on the fixture's tool run and the recorded output becomes `None`, so the 500 vanishes and the node grades clean. That is exactly the failure the tool rules exist to catch, arriving as a dropped payload rather than a wrong verdict. The fallback costs nothing on the wrapped shape (your test still passes unchanged) and there's a guard that fails without it.

Your "Left out" note flags tool *errors* from a real export as untested against live data — agreed, and that's the other half of the same assumption. Worth pinning whenever a real export is available.

**To sync locally:**

```
git fetch origin && git checkout s4-tool-child-runs
git reset --hard origin/s4-tool-child-runs
```

Check `git log origin/s4-tool-child-runs..s4-tool-child-runs` first if you have unpushed work. #103 (S-5) and #104 (S-6) are drafts, so I've left them alone — flip them to ready when you want them looked at.

PR #102, opened 2026-09-14
test(recorder): preserve existing inspection messages

MinhaKim02:
## What & why

Follow-up to #100.

This adds the non-blocking regression case noted during review: a contextual reason should append without erasing a structural or tool message that already exists on the origin step.

The integration test:

- creates a pre-existing `empty_output` tool-failure message;
- adds a contextual reason for the missing `docs` field;
- verifies both messages survive and the contextual reason is appended after the existing message.

This is a test-only change; production behavior is unchanged.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / internal
- [ ] Docs
- [x] Other: Regression test

## Checklist

- [x] I have joined the [ARGUS Discord](https://discord.gg/67XTFTDSgd).
- [x] I have signed the [CLA](../CLA.md).
- [x] My change targets the **open-source core** only.
- [ ] Tests added/updated and `pytest` passes locally.
- [ ] `ruff check .` passes.
- [x] Docs updated if behavior or the public API changed (not applicable: test-only).
- [ ] PR has a label.

## Notes for reviewers

The new assertion would fail if `_blame_origins` returned to overwriting `inspection.message`, while leaving the existing production implementation untouched.


VaradDurge:
Thanks for picking this up — it's the exact gap I flagged on #100, and the half worth pinning.

I rebased onto current `pivot/fat-traces` (the base moved for #89, #87 and #101) and had to change the graph, because the test did not pass as written. Your authorship is preserved; new head is `81864da`.

**What was wrong, and it isn't a small detail.** The test used `prepare` returning `{}`. On that shape the message ends up as:

```
'Tool failures: empty_output on "_output"'
```

— no contextual reason at all, so `assert "`docs`" in message` fails. It fails the same way on your own base (`87d7706`), so it was never green; your checklist has *"Tests added/updated and `pytest` passes locally"* and *"`ruff check .` passes"* unticked, which lines up. Worth running both before pushing — CI on this repo needs a maintainer to release workflows for first-time fork contributors, so a red test can sit invisible for a while.

**Why that shape can never work.** `contextual` deliberately stays quiet when the step it would blame returned `{}` — it defers to `empty_output` so one no-op isn't reported twice. So an empty update yields the tool message alone and there is nothing to append. The scenario you wanted is unreachable through a node that returns nothing.

**The shape that does reach it:** an origin with a *non-empty* update that still misses the consumed field. `prepare` now swallows an upstream error into an `error` key and never writes `docs`:

```
'Tool failures: error_response on "error"; Field `docs` is read by `answer` but no step wrote it.'
```

Two layers on one step, tool message first, contextual appended — which is precisely what you set out to assert, and the assertions are otherwise yours. I added the reachability reasoning to the docstring so the next person doesn't "simplify" it back to `{}`, and failure messages that print the actual string.

**Verified:** restoring the overwriting one-liner from #100 fails this test *and* `test_contextual_reasons_accumulate_on_the_same_origin`, so it's a real guard rather than a restatement. Full suite 1042 passed / 4 skipped / 2 xfailed; both matrices 68; ruff clean; CI green on 3.9/3.11/3.12.

**To sync locally:**

```
git fetch origin && git checkout test/preserve-existing-inspection-message
git reset --hard origin/test/preserve-existing-inspection-message
```

PR #103, opened 2026-09-14
S-5: argus edges export and --edges on ingest

abhisheksharma001:
**Stacked on #101 (S-4).** This branch sits on S-4 rebased onto 87d7706, so the first three commits are S-4's. Review only the last two (a342eae, 9d0ae64). Draft until #101 lands.

## What changed
- `argus edges module:factory --out edges.json` compiles the graph once and writes the topology the recorder reads at attach (`recorder._topology`): `edge_map`, `conditional_sources`, `node_names`, `subgraph_parents`.
- `argus ingest langsmith <file> --edges edges.json` uses those edges instead of the step-order guess, and drops subgraph parents by name instead of guessing from nesting. `--edges` is optional; without it nothing changes.
- A missing or malformed edges file exits 2 and saves nothing.
- `langgraph` is imported only in `cmd_edges.py`; `ingest/` still imports without it (existing test).

## Proved
- With `--edges` from the demo graph, the saved `graph_edge_map` equals a live recorder run's, and `check last` still fails `summarize`.
- The demo graph is a straight line, so that test passes even without the wiring. A second test feeds an edge map step order can never produce and checks it is saved.
- Break proofs: using the step-order guess with `--edges` fails exactly `test_the_edges_file_replaces_the_step_order_guess`; ignoring `subgraph_parents` fails exactly `test_edges_name_the_subgraph_parents_instead_of_nesting`.
- Full suite: 1029 passed. ruff clean. No LLM call in any test.

## Left out
- No change to how edges are guessed without `--edges`.
- No subgraph fixture exported from LangSmith; subgraph-by-name is covered by a unit test on `node_runs`.


VaradDurge:
#101 has landed, so I rebased this onto current `pivot/fat-traces` (dropping S-4's three commits) and reviewed it. New head is `2916415`. CI green on 3.9/3.11/3.12, full suite 1048 passed, both matrices 68. Leaving it in draft — flip it when you're ready.

The design is right: opt-in, nothing changes without `--edges`, `langgraph` confined to `cmd_edges.py` (verified `argus.ingest.langsmith` still imports with langgraph and langchain_core blocked), and the export reads topology through `recorder._topology` so an ingested run gets exactly what a live one would.

**Verified by hand:** `argus edges demo_graph:build_app --out edges.json` writes the expected map; ingest with it still blames `summarize` and `check last` exits 1; malformed file, missing file, and a bad factory spec all exit 2 with nothing saved.

**Two things I fixed on top** (`2916415`):

**1. The second break proof does not hold.** You list "ignoring `subgraph_parents` fails exactly `test_edges_name_the_subgraph_parents_instead_of_nesting`". It doesn't — I replaced `node_runs(runs, set(edges["subgraph_parents"]))` with `node_runs(runs)` in `ingest_langsmith` and the whole file still passed 11/11. That test calls `node_runs(runs, parents)` **directly**, so it never exercises the wiring; ingest could stop passing the file's parents entirely and CI would stay green. Added a test that drives it through the CLI with an edges file naming `summarize` a subgraph parent, asserting it does not become a step. That one does fail when the call site is broken.

**2. A wrong or stale edges file silently turns a failure into a pass** — the more serious one. The file only has to be *well-formed*, not *right*. Point ingest at a map describing another graph:

```
steps : [('search','pass'), ('summarize','pass'), ('answer','pass')]
first_failure: None        exit 0
```

`summarize` has no successors in that map, so `empty_output` cannot fire and the silent node this fixture exists to catch grades **clean**. Without `--edges` at all, the same trace correctly blames it — so passing a stale file is strictly worse than passing none, which is the wrong way round for an opt-in flag. It is easy to hit for real: refactor the graph, forget to re-export `edges.json`, and CI goes green on a broken pipeline.

Ingest now refuses when the trace ran nodes the file does not describe — exit 2, nothing saved, message naming the nodes and pointing at `argus edges`. Both new guards fail when their mechanism is reverted.

Your "Left out" note says the straight-line demo can't prove the edges wiring, which is exactly right and is why `test_the_edges_file_replaces_the_step_order_guess` feeding an impossible map was the correct call. The same reasoning is what the two additions extend.

**To sync locally:**

```
git fetch origin && git checkout s5-edges-sidecar
git reset --hard origin/s5-edges-sidecar
```

#104 (S-6) is stacked on this and will need the same rebase once this lands.

PR #104, opened 2026-09-14
S-6: a skinny LangSmith trace refuses instead of passing

abhisheksharma001:
**Stacked on #103 (S-5), which carries #101 (S-4).** Only the last two commits are S-6.

## What changed
- `ingest_langsmith` refuses before grading with `IncompleteTraceError` when the root run's `outputs` are absent or `{}` (what LangSmith `hide_outputs=True` produces), or a node run has no `inputs`. The CLI exits 2 and `.argus/runs/` gets no file.
- One node's `{}` outputs with root outputs present is still graded as a silent node (BUG-1), so the demo fixture still blames `summarize`.

## Proved
- `pytest tests/test_ingest_langsmith.py -q -k skinny`: 4 pass.
- Breaking each guard fails exactly its test: root check → `test_skinny_trace_with_outputs_hidden_saves_nothing`; inputs check → `test_skinny_node_run_without_inputs_saves_nothing`; refusing any node's `{}` → `test_not_skinny_when_only_one_node_has_empty_outputs` (+4 others).
- Full suite 1033 passed; ruff clean. No LLM call.

## Left out
- Zero node runs: already refused by `grading.finish`; a test pins it, no new code.
- `inputs == {}` (hide inputs) is still graded.


VaradDurge:
Merged as 0a1098c on `pivot/fat-traces`. The PR itself conflicted because it still carried the pre-squash S-4/S-5 commits, so the two S-6 commits were cherry-picked and squashed with authorship preserved. Tests green (85 across ingest + both matrices), ruff clean.

PR #105, opened 2026-09-15
S-7: --consumers on ingest wires the contextual layer

abhisheksharma001:
## What changed

- `argus ingest langsmith --consumers consumers.json` reads a `{"field": ["reader", ...]}` map and passes it to `grading.finish`. `ingest_langsmith` already accepted `consumers=` since S-3, so this is the CLI flag plus a loader.
- `load_consumers` (in `cmd_ingest.py`) checks the shape. A bare-string reader such as `{"customer_id": "respond"}` would otherwise be iterated letter by letter, match no step and grade clean. A wrong-shape file exits 2 and saves nothing.
- New fixture `tests/fixtures/langsmith/drop_graph.jsonl`, generated by `scripts/make_langsmith_fixture.py --drop`. It is a four-node graph where `enrich` nulls `customer_id` and `respond` reads it two steps later.

## Proved

- Acceptance: ingesting `drop_graph.jsonl` with `{"customer_id": ["respond"]}` makes `argus check last` exit 1 with `first_failure_step == "enrich"` and a critical `missing_field` naming `respond`.
- Without `--consumers` the same file does not fail (two `enrich` warnings, no `missing_field`). The declared reader is what finds the drop.
- Break proofs:
  - Not passing the map to ingest fails `test_consumers_blame_the_step_that_dropped_the_field`.
  - Skipping the shape check fails `test_consumers_file_of_the_wrong_shape_saves_nothing`.
  - Both restored.
- Full suite: 1051 passed, 4 skipped, 2 xfailed. `ruff check` is clean.

```
PYTHONPATH=src pytest tests/test_ingest_langsmith.py -q -k consumers
```

## Left out

- No inferring consumers from source; that is a later step.
- Branched off `pivot/fat-traces` at 8cfec5b and independent of S-6 (#104). Both add tests before the same anchor in `tests/test_ingest_langsmith.py` and a note to the PRD, so whichever merges second needs a trivial rebase.


VaradDurge:
Merged as cc348d5 on `pivot/fat-traces`. Cherry-picked on top of S-6 (#104); the expected conflict in `tests/test_ingest_langsmith.py` was both PRs adding tests at the same anchor — resolved by keeping both blocks. 88 tests green across ingest + both matrices, ruff clean.

PR #106, opened 2026-09-15
S-8: LLM rows carry token usage and warn on truncation

abhisheksharma001:
## What changed

Step S-8 of `docs/prd_abhishek.md`.

- **Recorder:** `ArgusRecorder.on_llm_end` files each model call under the node step it ran in. It walks chain parents, so a call inside `prompt | llm` still finds its step.
- **Ingest:** `argus ingest langsmith` reads `run_type == "llm"` child runs (`llm_calls_by_step`).
- **Shared:**
  - Both paths fill the existing `NodeEvent.llm_usage`, and `LLMCallInfo` gains `finish_reason`.
  - One parser, `llm_tracker.call_from_llm_outputs`, reads the tracer's `LLMResult` dict. The recorder rebuilds that shape with `dumpd`.
  - `session.on_node_end` adds a **warning** `truncated_llm_output` anomaly when a call's `finish_reason` is `length` or `max_tokens`. A truncated run still passes `argus check`.
- **Storage:** `load_run` now reads `total_llm_calls`, `total_tokens` and `total_cost_usd` back. They were saved but dropped on load, so every reloaded run showed 0 tokens (BUG-2 in the PRD).
- **Fixture:** `scripts/make_langsmith_fixture.py --llm` writes `tests/fixtures/langsmith/llm_graph.jsonl`. It uses a scripted chat model: no provider, no key, no network.

## Proved

- `pytest tests/test_recorder.py tests/test_ingest_langsmith.py -k "llm or trunc"`: 2 passed. On both paths `write` carries one warning `truncated_llm_output` and `total_tokens == 60`.
- Full suite: 1050 passed, 4 skipped, 2 xfailed. `ruff check` is clean.
- Break tests (each restored afterwards):
  - Removing the parent walk fails only the recorder test.
  - Dropping `llm_usage` in ingest fails only the ingest test.
  - Removing the signal, not loading the totals, or making it `critical` each fail both tests.

## Deliberately left out

- **Spec deviation:** no new `NodeEvent.llm_calls` field, because `llm_usage` already holds the calls.
- **Per-step reload:** a reloaded step's `llm_usage` is still `None`. That is queued as S-12.
- **Scope:** no change to the old wrap path's `llm_tracker` handler, and no real model calls.


VaradDurge:
Merged as `786cc8a` on `pivot/fat-traces` (cherry-picked with authorship preserved; the PR branch predated the S-6/S-7 merges, so the button would have conflicted).

PR #107, opened 2026-09-15
S-12: a reloaded step keeps its llm_usage

abhisheksharma001:
**Stacked on #106 (S-8).** Until #106 merges this diff also shows the S-8 commits; the S-12 change is the last two commits (`63e3268`, `801644a`).

## What changed
- `storage._deserialize_event` now rebuilds `NodeEvent.llm_usage` (`LLMUsage` with `LLMCallInfo` calls). Before, a step loaded with `load_run` had `llm_usage is None` even though the saved JSON carried its calls, tokens and `finish_reason` (BUG-2, second half; S-8 fixed the run totals).
- A step saved without `llm_usage` still loads as `None`. Older calls without `finish_reason` load with `None` (dataclass default).
- New test `test_a_reloaded_step_keeps_its_llm_calls`: ingest `llm_graph.jsonl`, reload, and the `write` step has `finish_reason == "length"` and `total_tokens == 40`.
- PRD: S-12 done note; BUG-2 marked fixed.

## Proved
- `pytest tests/test_ingest_langsmith.py -k llm`: 2 passed. Full suite: 1051 passed, 4 skipped, 2 xfailed. `ruff check` clean.
- Break proofs (restored after each): not passing `llm_usage` to `NodeEvent` fails only the new test; keeping calls as raw dicts fails only the new test (`'dict' object has no attribute 'finish_reason'`).

## Left out
- What is saved is unchanged. `cmd_open_ui.py` untouched.
- `tool_calls` still reload as dicts, which matches the `list[dict]` type on `NodeEvent`.


VaradDurge:
Merged as `8127d63` on `pivot/fat-traces` (cherry-picked with authorship preserved; the PR branch predated the S-6/S-7 merges, so the button would have conflicted).

PR #108, opened 2026-09-15
S-10: pin langgraph>=0.6 and test the floor and 1.x in CI

abhisheksharma001:
## What changed

- `pyproject.toml`: `langgraph>=0.2.0` → `langgraph>=0.6` (default deps and the `[langgraph]` extra). The recorder path needs 0.6+: on 0.2.74, 56 pivot tests fail before the recorder runs.
- `.github/workflows/ci.yml`: new `langgraph` matrix axis, `~=0.6.0` and `>=1,<2`, installed in the same `pip install` as `.[dev]` so the resolver honours it. The 3.9 × 1.x leg is excluded because langgraph 1.x requires Python 3.10+. Five legs in total, each running the whole suite.

## What was proved (locally, Python 3.12)

- langgraph 0.6.11 (`~=0.6.0`): 1048 passed, 4 skipped, 2 xfailed.
- langgraph 1.2.11: 1048 passed, 4 skipped, 2 xfailed.
- `pip install -e . "langgraph==0.2.74"` fails with `ResolutionImpossible` (uv reports the same, unsatisfiable).
- Break proof: the same install against the old `>=0.2.0` floor resolves and would install 0.2.74, so the floor is what refuses.
- `ruff check .` clean. No test changed.

## Deliberately left out

- Exact `==0.6.0` pin in CI: `~=0.6.0` resolves to the latest 0.6.x (0.6.11).
- Dropping Python 3.9 support.
- Any change to tests or `src/`.

Independent of #104–#107; based on `pivot/fat-traces` 8cfec5b.


VaradDurge:
Merged as `1ea18e2` on `pivot/fat-traces` (cherry-picked with authorship preserved; the PR branch predated the S-6/S-7 merges, so the button would have conflicted).

PR #109, opened 2026-09-15
S-11: a crashed node run from a LangSmith file keeps crash blame

abhisheksharma001:
## What changed

- `scripts/make_langsmith_fixture.py --crash` traces a new graph under the stub LangSmith client. `lookup` writes `{"policy": {}}`, a bystander `audit` writes `notes`, and `price` raises on `state["policy"]["number"]`.
- `tests/fixtures/langsmith/crash_graph.jsonl` is the generated trace (4 runs).
- `tests/test_ingest_langsmith.py::test_a_crash_from_the_file_blames_whoever_wrote_the_empty_container` ingests the fixture and asserts that `check last` exits 1 with `first_failure_step == "lookup"`. It also requires a `missing_field` finding on `lookup` that names `number` and `price`, and no critical finding on `audit`.

No change to `inspector.py` or `ingest/`. The recorded `error` string already contains `KeyError: 'number'`, so S-3's pass-through was enough (the PRD stop rule did not trigger).

## Beyond the spec, and why

1. **Bystander `audit` node.** With only `lookup → price`, the fallback backward walk also blames `lookup`, so the test could not tell nested-container blame from luck.
2. **Path scrub.** Traceback text in `error` carried absolute host paths. The generator now rewrites them to `<site-packages>` and `<repo>`.

## Proved

- Break A: set `exc = None` in `ingest/langsmith.py` (the error is dropped) and the test fails. A first version that asserted only `first_failure_step` survived this break, because BA-006 (empty output) also flags `lookup`. The `missing_field` assertion closes that gap.
- Break B: set `nested = None` in `crash_origins` and the test fails, with blame moving to `audit`.
- Each break fails only this test.
- Full suite: 1047 passed, 6 skipped (environment-only skips), 2 xfailed. Ruff clean.

## Left out

- No change to crash-blame rules.
- No traceback parsing in `ingest/`.

Independent of #104–#108. It shares the test-file anchor region and PRD with S-6/S-7/S-8, so a later merge may need a trivial rebase.


VaradDurge:
Merged as `bb7ba30` on `pivot/fat-traces` (cherry-picked with authorship preserved; the PR branch predated the S-6/S-7 merges, so the button would have conflicted).

